PR #61, opened 2014-05-17
ObjectDisposedException from singleton factory

ivan-danilov:


ivan-danilov:
Any progress with this issue?

IanYates:
This seems related to #100 if I've read through it properly.  That is, the container, whether via a typed factory or not, is being asked to resolve an instance of something whilst constructing something else.  In that case, a relationship is created automatically between the resolved instances.  This could lead to double-disposal and perhaps the solution, at least for the typed-factory case, is for the typed-factory to know it's being asked to resolve in this manner and specifically avoid the double-disposal.

ivan-danilov:
@IanYates in test [here](https://stackoverflow.com/questions/23706455/why-castle-windsors-typed-factory-registered-as-singleton-lifestyle-is-disposed) resolves are invoked one by one, but the issue still happens.
The problem is not so much with double-disposals, but rather with disposing singleton-lifestyle objects which are still in use.

ivan-danilov:
@kkozmic @IanYates could you please at least direct me where it should be fixed, may be I could work it out myself? I have an impression this issue was just forgotten, and that is sad...

IanYates:
@ivan-danilov - I don't think it's been forgotten but the devs who really know the guts of Castle are busy with other projects, etc.  It'd probably also help keep it visible if an issue was made rather than a PR - as @kkozmic said earlier, once this PR is pulled in to get the failing tests into the repository then the PR will be closed (although at least automated tests will fail).

I didn't see your comment from 19 days ago sorry.  My comment from 20 days ago is, I suppose, more related to @xlegalles problem rather than the core issue here although they're all in a similar area as far as calculation and releasing of burdens are concerned.

I'm not that familiar with the Castle internals to give you a good indication of where to go digging I'm afraid.

petr-k:
Is there a known workaround for this?

johnEsacks:
Noticed this PR has been out there for a while.  I have the same issue.  Has this been fixed in master and/or is this PR usable in its current state as a workaround? If not, what is the current known workaround?

jnm2:
Hey, I need this too.

hammett:
If double disposal is the underlying issue, we should just adhere to the .net fx guidelines: calling dispose multiple times is allowed, and dispose should never throw.

jnm2:
My problem is not double-disposal but disposal too soon. Created #139 explicitly explaining it.

ivan-danilov:
@hammett no, it is not double dispose issue in general. In fact there're examples in the thread where object should be never disposed even once, but it is unfortunately. It bites.

hammett:
@ivan-danilov Will the PR fix the issue?

ivan-danilov:
@hammett No, it is just unit tests that show the issue. I mean, they're failing and should be not as soon as the issue is fixed. I don't know Windsor's internals good enough to provide a fix right now.

hammett:
Ok, got it.

bindu-purhar:
@hammett Any progress on this issue?

hammett:
Nope. thanks for the bump. I'll take a look on the weekend.

bindu-purhar:
Any luck?

hammett:
I apologize, but I still couldn't find time to investigate this.

jnm2:
I've also had ongoing problems and am interested in this.

bindu-purhar:
I sure am seeing this crash a lot.  Hope you can get to it soon.

InnerException = System.ObjectDisposedException: The factory was disposed and can no longer be used.
Object name: 'this'.
   at Castle.Facilities.TypedFactory.Internal.TypedFactoryInterceptor.Intercept(IInvocation invocation)
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at Castle.Proxies.IGlobalResourceDictionaryFactoryProxy.Create()

hammett:
Could you guys post the a snippet of code using the TF that is throwing this exception? Also, are you guys using child containers?

hammett:
Never mind, forgot this was a PR with tests/repro included. Looking into it now...

bindu-purhar:
Any luck?

hammett:
No, this is super hard. It's a TypedFactory within a TypedFactory, and I'm not even sure this scenario is officially supported. Debugged for a couple of hours and there's a long road ahead...

ivan-danilov:
@hammett hm, in [this](https://github.com/castleproject/Windsor/pull/61/commits/aa802f05943a863764c7193b5bc3fac026ffb7bf) case there's only one typed factory, and the issue still reproduced. Seems that factory incorrectly tracks its products lifetime. Could you please share your investigation results?

hammett:
From your link:

> IGenericFactory<IGenericService<int>>>

Like I said I'm not even sure that's a supported scenario. It seems to be a issue with the component burden integration with TypedFactory facility, since I havent really written that code, there's a lot to absorb.

ivan-danilov:
Actually I tried to use classes that were available in Windsor's codebase (`IGenericFactory<T>`) and added `IGenericService` by the similar logic. I don't think it is important, but will try to make simpler case with less moving parts.

hammett:
I dont think a simpler case is needed, the issue is already highlighted by your test cases. I appreciate the contribution.

ivan-danilov:
So... what do you mean by unsupported case then? Types registered as singletons returned from TypedFactory instance?

hammett:
A recursive relationship between types implemented by the TypedFactory facility.

jonorossi:
I'm going to close this pull request as we are aware the unit tests are here, and will keep track of this issue in #134.

Issue #75, opened 2015-01-13
Resolve component implementing single generic interface throws ArgumentException


ezsilmar:
I'd like to resolve component which implements generic interface with different generic parameters. My code:

``` c#
    public interface IOuterService<T, U> : ICanBeResolvedByContainer
    {
    }

    public interface IWrap1<T> : ICanBeResolvedByContainer
    {
    }

    public interface IWrap2<T> : ICanBeResolvedByContainer
    {
    }

    public interface IService1<T, U> : IOuterService<IWrap1<T>, U>
    {
    }

    public interface IService2<T, U> : IOuterService<IWrap2<T>, U>
    {
    }

    public class Component<T, U> : IService1<T, U>, IService2<T, U>
    {
    }
```

Then let's register everything which implements ICanBeResolvedByContainer:

``` c#
    var container = new WindsorContainer();
    container.Register(
        Classes
            .FromAssemblyContaining<ICanBeResolvedByContainer>()
            .BasedOn<ICanBeResolvedByContainer>()
            .WithServiceFromInterface(typeof (ICanBeResolvedByContainer))
            .WithServiceSelf());
    var dep = container.Resolve<Component<int, int>>();
```

The last line throws ArgumentException: An item with the same key has already been added.

Expected behavior: resolve or throw meaningful exception.

My investigations lead to problem in [DefaultGenericHandler.EnsureInterfaceMappingInitialized](https://github.com/castleproject/Windsor/blob/016730de012f15985410fb33e2eb907690fe5a28/src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs#L472). GetGenericTypeDefinition is used here as dictionary key, and there are obviously two similar keys in this case: typeof(IOuterService<,>). 

Is this just a not implemented case?

kkozmic:
Without running your test or looking at any code, it looks like it might be a bug. Part of the problem appears to be related to the fact that semi-open generic types are involved, but ideally it should just work.

dalefrancis88:
Hey,

I created a fix for this. https://github.com/castleproject/Windsor/pull/81

ezsilmar:
dalefrancis88, thank you for the fix.

I've tried it and found this case (setup is the same):

```c#
   Container.Resolve<IOuterService<IWrap1<int>, int>>()
```

[...]

PR #81, opened 2015-04-07
Issue 75 Fix

dalefrancis88:
Updated code to not use dictionary. It now looks through interfaces for
the expected service type.

I think this may have very minor performance difference but not sure
there is enough that it matters

Not sure what is going on with DefaultGenericHandler.cs commit. Let me know if i have to do something different


jonorossi:
@dalefrancis88 there is a line ending issue with your PR, could you please take a look at it.

kkozmic:
This is a fix for #75

PR #90, opened 2015-06-24
Remove obsolete docs for typed factory facility

### LINKED ISSUE #82: Wrong documentation for typed factory facility
[The documantation for "Get" methoid behavior](http://docs.castleproject.org/Windsor.Typed-Factory-Facility-interface-based-factories.ashx#Get_methods_lookup_components_by_name_4) is misleading. It still says that the default behavior is to fallback to type resolution if the get by name fails.

Apperantly this have been changed with commit 213d45dbaca7b3e0cd2ce06677062974311dd589 and is no longer valid for v3 and up.

sloncho:
fixes #82


jonorossi:
Thanks

PR #91, opened 2015-06-24
Update basic-tutorial.md

petersmax:
Point references to their proper namespaces and classes.


kkozmic-seek:
I'd argue that whole tutorial is lacking in quality... Would be nice to give it a proper upgrade.

petersmax:
Agreed but I learned some very basic working stuff so I wouldn't knock it too hard

jonorossi:
@petersmax thanks for the contribution, however I'm not sure why we'd want the tutorial to fully qualify the namespaces of all classes, to me it just makes it look more complicated?

petersmax:
well naturally I would have to agree in that case


____________
Maximus Peters

petersmax:
but would like to note that the namespace references to Component is
ambiguous and it keeps thinking that it is referencing .NET .Component
namespace... as for ClassLibrary it also could not find it either until I
added that.


____________
Maximus Peters

PR #116, opened 2016-01-07
Fix case sensitivity issue that can cause UsingFactoryMethod to fail …

dohansen:
…if 2 class names only differ by case.

We use an obfuscator that will generate class names that only differ by case. This was causing an issue if one of these classes was registered with UsingFactoryMethod. A class of type FactoryMethodActivator&lt;T&gt; is created with the wrong type parameter because it will create a class of FactoryMethodActivator&lt;t&gt; instead which caused the registration to fail.

### DIFF
--- a/src/Castle.Windsor.Tests/TypeNameConverterTestCase.cs
+++ b/src/Castle.Windsor.Tests/TypeNameConverterTestCase.cs
@@ -262,5 +262,22 @@ namespace CastleTests
 
 			Assert.AreEqual(message, exception.Message);
 		}
+
+		class TestCaseSensitivity { }
+		class TESTCASESENSITIVITY { }
+
+		[Test]
+		public void Can_resolve_exact_match_if_two_classes_exist_that_differ_only_by_case()
+		{
+			var type = typeof(IGeneric<TestCaseSensitivity>);
+			var name = type.AssemblyQualifiedName;
+			var result = converter.PerformConversion(name, typeof(Type));
+			Assert.AreEqual(type, result);
+
+			var type2 = typeof(IGeneric<TESTCASESENSITIVITY>);
+			var name2 = type2.AssemblyQualifiedName;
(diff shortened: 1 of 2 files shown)


jonorossi:
@dohansen you've got some line ending issues with your commit.

dohansen:
When I pulled down the repo, git reported lots of changed files. I had core.autocrlf set to false. I was able to get rid of the modified files by using  core.autocrlf  set to input and clearing and checking the files out again. One of the files seems OK, but the other one I couldn't get to commit without reporting line ending changes.


Do you have any advice on how to fix it?


________________________________

jonorossi:
@dohansen it is annoying we've still got an unnormalized repo, I'll have to look into that. I think I've worked around that problem in the past using `core.autocrlf true` assuming you are on Windows. It looks like a combination of an unnormalized repo and the fact we've got a .gitattributes file now.

dohansen:
I am on Windows. The only way I can get it to work is to first commit the line ending changes to src/Castle.Windsor.Tests/TypeNameConverterTestCase.cs as one commit and then apply my changes after that. It seems that what is currently committed in the repo is in conflict with the settings in .gitattributes so it will always update the line endings on commit no matter what core.autocrlf is set to.

Toxantron:
Came to the same conclusion. I think what might work is to remove all eol-settings from .gitattributes without committing them. I didn't have time to try it with my own PR but since .gitattributes overrides any settings like ``--global core.autocrlf true/input`` it could bring back expected behaviour.

jonorossi:
I've applied your patch as ff25c560171ec2ce0bb53d9bcdd584b702b86ba1 with the whitespace fixed.

@dohansen you should fix your `git config user.email` even if you want to keep your email private. I've amended this commit from `username@users.noreply.github.com` to `dohansen@users.noreply.github.com` so it gets attributed to you.

PR #125, opened 2016-03-02
Fixed syntax error in example code.

bobtoebes:
Added missing parenthesis to overloaded For and ImplementedBy example code.


jonorossi:
Thanks

PR #128, opened 2016-03-26
All implicit object casts fixed. Fixes #127

### LINKED ISSUE #127: Implicit cast converts null to object
In [ServiceOverride.cs L82](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/ServiceOverride.cs#L82) the implicit cast converts a ServiceOverride of null into an object. This took me a while to figure out because it is behavior I never saw before. Is this intended?
````c#
public static implicit operator Dependency(ServiceOverride item)
{
    return new Dependency(item);
}
````

I would have expected the cast like this:
````c#
public static implicit operator Dependency(ServiceOverride item)
{
    return item == null ? null : new Dependency(item);
}
````

Toxantron:
As described in #127 this makes sure that imlicit casts do not create objects out of thin air.

### DIFF
--- a/src/Castle.Facilities.WcfIntegration/Behaviors/Security/FetchCertificate.cs
+++ b/src/Castle.Facilities.WcfIntegration/Behaviors/Security/FetchCertificate.cs
@@ -23,7 +23,7 @@ namespace Castle.Facilities.WcfIntegration.Behaviors
 
 		public static implicit operator AbstractCredentials(FetchCertificate finder)
 		{
-			return new CertificateCredentials(finder);
+			return finder == null ? null : new CertificateCredentials(finder);
 		}
 	}
 }
--- a/src/Castle.Facilities.WcfIntegration/Behaviors/Security/FetchCertificateBySubject.cs
+++ b/src/Castle.Facilities.WcfIntegration/Behaviors/Security/FetchCertificateBySubject.cs
@@ -20,7 +20,7 @@ namespace Castle.Facilities.WcfIntegration.Behaviors
 
 		public static implicit operator AbstractCredentials(FetchCertificateBySubject finder)
 		{
-			return new CertificateCredentials(finder);
+			return finder == null ? null : new CertificateCredentials(finder);
 		}
 	}
 }
--- a/src/Castle.Windsor/MicroKernel/Registration/Parameter.cs
+++ b/src/Castle.Windsor/MicroKernel/Registration/Parameter.cs
@@ -74,7 +74,7 @@ namespace Castle.MicroKernel.Registration
 
 		public static implicit operator Dependency(Parameter parameter)
 		{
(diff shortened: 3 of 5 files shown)


Toxantron:
Sorry, line endings broken.

kkozmic:
Yeah I noticed. No need to close the PR, just force push if you normalise it. Perhaps @jonorossi has some context on what line endings settings should be these days. I remember we never normalised it, back in a day, and that sadly might be what you're experiencing.

Toxantron:
Seriously? I have been trying to get this to work for hours and still ended up with broken line endings.

kkozmic:
Yeah, I hear ya.

@jonorossi have we tried something like this to get that resolved once and for all? http://superuser.com/questions/293941/rewrite-git-history-to-replace-all-crlf-to-lf

Toxantron:
Old doc said CRLF was to be used and some does .gitattributes. But it seems as if some files already have LF, which might explain why my git keeps trying to fix it.

However more common OSS style is LF in repo and CRLF on Windows only.

Toxantron:
Besides I must work somehow, I probably just missed something. Thanks for taking the time to look into it anyway. 
I'll get back to this tomorrow.

jonorossi:
I've been using `core.autocrlf true` without problem in Windsor.

I've got normalising line endings once and for all (but not as far as rewriting all history) on my todo list for Castle Core v4 after we get rid of branches: https://github.com/castleproject/Core/issues/111. My plan was to do the same with Windsor, and replace that poor .gitattributes file with something sane.

Toxantron:
I'll try again with autocrlf later. Thanks for the feedback. What EOL do you plan? LF or CRLF?

jonorossi:
@Toxantron if you can't get it working, let me know. I'll make the changes and attribute them to you.

LF normalised object database (i.e. `* text=auto`) for everything, nothing explicitly set.

Toxantron:
I didn't have much time over Easter, but we are not in a hurry I assume. However if you got time and a working repo feel free to do it. It's just the 5 occurrences of ``public static implicit`` that I replaced with ``item == null ? null : new Dependency(item)``. Otherwise I suppose I will find time in the next couple of days to work around this. I want to set up a working dev enviroment for Castle anyway because I think there might be another bug as well as porting what we build on top into a facility.

If I understand git correctly the ``.gitattributes`` file overrides ``core.autocrlf true`` anyway and if LF is what you want to go for the old doc page maybe should be removed. ;-)
[Old Castle documentation about git](http://docs.castleproject.org/(S(bcwix455ivqyod45twfzyj45))/History.aspx?Page=How-to-submit-a-fix-to-any-Castle-Project&Revision=9)

jonorossi:
I've made the change, see #127.

Issue #101, opened 2015-09-24
NHibernateIntegration facility


fsalas:
Hi, by any chance is the facility broken for castle 3, I'm trying to use it, and keep getting an exception, or the facility is deprecated and I should use another in replacement as the nuget package Castle.Facilities.NHibernate

I'm building some tests with the following code, and getting 

```
System.MissingMethodException : Method not found: 'Void Castle.Core.DependencyModel..ctor(Castle.Core.DependencyType, System.String, System.Type, Boolean)
```

The Stack Trace is:

```
   at Castle.Facilities.NHibernateIntegration.Internal.NHSessionComponentInspector.ProcessModel(IKernel kernel, ComponentModel model)
   at Castle.MicroKernel.ModelBuilder.DefaultComponentModelBuilder.<>c__DisplayClass6.<BuildModel>b__4(IContributeComponentModelConstruction c)
   at System.Collections.Generic.List`1.ForEach(Action`1 action)
   at Castle.MicroKernel.ModelBuilder.DefaultComponentModelBuilder.BuildModel(IComponentModelDescriptor[] customContributors)
   at Castle.MicroKernel.Registration.ComponentRegistration`1.Castle.MicroKernel.Registration.IRegistration.Register(IKernelInternal kernel)
   at Castle.MicroKernel.DefaultKernel.Register(IRegistration[] registrations)
   at Castle.Facilities.NHibernateIntegration.NHibernateFacility.RegisterDefaultConfigurationBuilder()
   at Castle.Facilities.NHibernateIntegration.NHibernateFacility.RegisterComponents()
   at Castle.Facilities.NHibernateIntegration.NHibernateFacility.Init()
   at Castle.MicroKernel.Facilities.AbstractFacility.Castle.MicroKernel.IFacility.Init(IKernel kernel, IConfiguration facilityConfig)
   at Castle.MicroKernel.DefaultKernel.AddFacility(String key, IFacility facility)
   at Taxo.Framework.Test.Persistance.NHibernate.NHibernateModuleInitialization.ConfigureNHibernateFacility(IWindsorContainer _ioc) in C:\Proyectos\Sistemas\CatalogLegacy\dev\Framework\Test\Test\Persistance\NHibernate\NHibernateModuleInitialization.cs:line 74
[...]

PR #135, opened 2016-06-06
NHibernate facility repo name changed

janv8000:


jonorossi:
@janv8000 thanks for the pull request, however that repo is readonly as it is an old one before the facility was merged into the Windsor repo.

Alternatively I've fixed the link (6095db25ee4786269eeec894550365a060ec6f8a) to point to the doc page for the facility, and added a note to the facility doc page (349a2ab7d4c4b710b871fbe1f1a2eeedd736bd13) linking to #101.

Issue #146, opened 2016-09-02
Performance Counter Instances Never Removed


mackenzieajudd:
We ran into a "Custom counters file view is out of memory" exception in our application running in IIS which brought this issue to light.

We have performance counters enabled as instructed here: https://github.com/castleproject/Windsor/blob/master/docs/performance-counters.md

I found that the TrackedComponentsPerformanceCounterWrapper class never removes the PerformanceCounter instance that it wraps on disposal. If performance counters are enabled for an application which is restarted often the available performance counter memory can be exhausted because a new performance counter, with a unique name, is created every time on startup.

It would be nice if TrackedComponentsPerformanceCounterWrapper removed the wrapped performance counter instance on disposal.

To reproduce:
1. Create a console app with performance counters enabled (https://github.com/castleproject/Windsor/blob/master/docs/performance-counters.md)
2. Start and stop the app multiple times
3. Open Performance Monitor and look at the Castle Windsor category. You should see a performance counter for each time you started the app.

jonorossi:
@mackenzieajudd I'm don't think the intention of the Windsor performance counters was to be used in production always enabled, there shouldn't be any harm though other than this defect obviously. However, won't an application restart cause IIS to tear down the application pool process without really "shutting down" applications resulting in the same outcome?

The code currently calls `PerformanceCounter.Dispose` during clean up, a quick skim of the MSDN docs seems to indicate this is all that is required. What do you mean by remove the performance counter?

Any chance you could provide a pull request with the fix?

mackenzieajudd:
@jonorossi thanks for the quick response!

You may be right about the IIS teardown I'm not sure.

From what I can see `PerformanceCounter.Dispose` releases the resources it is using but won't remove a performance counter from the machine. When you create a new instance of `PerformanceCounter` it will either hook into an existing performance counter if the Category and Name already exist or create a new one if not.

In our case Windsor creates a new performance counter each time because its name is unique (Based on process name and id) to account for multiple instances being tracked at the same time. Because it's not re-using an existing performance counter instance I think the instance should be removed when Windsor is done with it so they don't pile up.

I'll put something together and hopefully have a pull request this week!

I've attached a picture to help illustrate what I'm talking about. The performance counters shown in it have all been created by Windsor but where not removed when the container was disposed and the application was closed.

[...]

PR #149, opened 2016-09-21
Fix for issue #146: Performance Counter Instances Never Removed

mackenzieajudd:
The fix turned out to be quite simple (thanks @jonorossi for the tips!), the InstanceLifetime of the created PerformanceCounter had to be set to Process. This causes the created performance counter instances to be removed when the process closes.

I also updated performance-counters.md to reflect this.

### DIFF
--- a/Changes.txt
+++ b/Changes.txt
@@ -975,3 +975,5 @@ Beta 1  - 21/jan/2005
 - Fixed: Remove method
 
 - Fixed: Windsor: Proxy for components with (service != impl)
+
+- Changed: Performance Counter InstanceLifetime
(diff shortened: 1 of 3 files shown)


kkozmic:
Nice one

Changes.txt
```diff
@@ -975,3 +975,5 @@ Beta 1  - 21/jan/2005
 - Fixed: Remove method
 
 - Fixed: Windsor: Proxy for components with (service != impl)
+
+- Changed: Performance Counter InstanceLifetime
```
jonorossi: Thanks for updating the changelog, however it is sorted with the newest releases at the top. Could you move this to the first line in the file, we add the release line once we ship.

jonorossi:
Looks good, fixes #146.

PR #157, opened 2016-11-24
Added ability to disable DelegateFactory for TypedFaciliy.

alkampfergit:
This modification allows the disabilitation of DelegateFactory, this imply
that, if an object has a dependency from Func<A> it gets registered with
TypedFactory only if someone explicitly register Func<A> with code

Container.Register(Component.For<Func<A>>().AsFactory());


fir3pho3nixx:
@alkampfergit - Is there an issue we can link this back to?

jonorossi:
@alkampfergit without an issue describing the problem this is attempting to fix I'm now going to close this pull request. Looking at the included unit test I wonder if making it facility wide is the right choice, per factory registration is probably where it belongs.

PR #169, opened 2017-02-09
All types failing test win10 vs2015 update 3 changelist

fir3pho3nixx:
The change list changes for PR: #168


jonorossi:
#168 hasn't been merged yet so why not just include the additions to the changelog in it?

It is also best not to predict the next version number, either leave it blank or put it under an "Unreleased" header.

fir3pho3nixx:
@jonorossi - I had a suspicion the PR would not go in directly. I split the two for flexibility mainly. Also was not sure about this change, so wanted to isolate it from the rest. 

Should I not touch that file?

jonorossi:
@fir3pho3nixx apologies if my comment was confusing. You should definitely update the changelog in the pull request where you make the changes. The comment here was about the new header you added "3.4.1 (TBD)", either don't include the header or name the header "Unreleased".

Issue #174, opened 2017-02-11
Prevent unit tests that require administrator privileges from running by default


jonorossi:
Following on from the discussion in #165, we decided that it would be worthwhile to investigate improving the developer experience to not require those developing Windsor to have to set up a bunch of environment stuff (e.g. MSDTC and WCF bindings) and run Visual Studio as an administator.

We should be able to take advantage of NUnit's categories to enable these "integration tests" in addition to the use of the `[Explicit]` attribute, we should probably create an `[IntegrationTest]` attribute (or something similar) in our test project so we don't have to duplicate string. The NUnit documentation indicates that enabling a category will run any test marked as explicit. This will allow the build server to run the full set of tests while those not working on the WCF or automatic transaction facilities don't have to fuss about with annoying environment set up.

/cc @fir3pho3nixx

fir3pho3nixx:
I manage to get a unit test only run with the following PR #175. 

To test it try `build NET40 RunUnitTests`.

Running just `build` still defaults to `build NET40 RunAllTests`.

`build NET40 RunUnitTests` passes in a cmd.exe window without `admin` privs.

jonorossi:
Thanks, closing.

PR #175, opened 2017-02-11
Categorise unit tests

fir3pho3nixx:
This was branched from `build-silverlight-zero`(just to make things easier).


fir3pho3nixx:
Think this closes #174

src/Castle.Windsor.Tests/Categories/UnitTestAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace CastleTests.Categories
+{
+	using NUnit.Framework;
+
+	public class UnitTestAttribute : CategoryAttribute
+	{
+		public UnitTestAttribute() : base("Unit")
```
jonorossi: Can you explain why we need 3 category attributes, I've done something similar before and don't understand why you'd need a `UnitTestAttribute`.

Castle.Windsor-SL.sln
```diff
@@ -1,68 +0,0 @@
-﻿
```
jonorossi: All the Silverlight changes merged in the other PR are still showing up here, maybe you just need to rebase to get GitHub to refresh the diff.

src/Castle.Windsor.Tests/Categories/UnitTestAttribute.cs
fir3pho3nixx: Just the opposite flavour which allows: 

nunit-console.exe /include:unit

As opposed to:

nunit-console.exe /exclude:Integration,Manual, AnyOtherConceivableCategoryInTheFuture

Not particularly tied down to it, very quick to remove if you want ...

fir3pho3nixx:
This has also been synced with the silverlight changes. I found some more statlight tooling in the directories and msbuild which I removed.

src/Castle.Windsor.Tests/Categories/UnitTestAttribute.cs
jonorossi: My design suggestion in #174 of only a single integration category and using `Explicit` for integration tests was for simplicity and familiarity for users, but importantly so that integration tests don't automatically get run by ReSharper or the VS test runner. I'd also much prefer not to have to remember to put an extra `UnitTest` attribute on classes, especially that `TestFixture` is no longer required in NUnit 2.5 so we can start omitting that attribute.

Can you explain the need for the `ManualTest` attribute, isn't that what explicit is for, which is why they are already marked as such?

fir3pho3nixx:
Ouch! Ouch ... closing it! Please see #178.

jonorossi:
> Ouch! Ouch ... closing it! Please see #178.

😄 . If you weren't already aware you can just update your branch including removing commits and it'll be updated in the pull request.

PR #178, opened 2017-02-13
Build unit vs integration

fir3pho3nixx:
A very simple turn key way running Castle Windsor tests without requiring admin privileges. 
    `build` <- still runs all tests and requires admin privs.
    `build NET40 RunUnitTests` <- Will exclude Wcf Integration tests which do not require ele. privs.

### DIFF
--- a/buildscripts/Build.proj
+++ b/buildscripts/Build.proj
@@ -236,7 +236,7 @@ limitations under the License.
 
 		<Message Text="Running tests from assemblies: $(TestAssemblies)" />
 		
-		<Exec Command="$(NUnitPath)$(NUnitExecutable) /nologo $(TestAssemblies) /exclude:IntegrationTest,Explicit /xml=$(NUnitTestResultXmlFiles) /framework=$(TargetFrameworkVersion)" Condition=" '$(TEAMCITY_VERSION)' == '' and '$(OS)' == 'Windows_NT' " />
+		<Exec Command="$(NUnitPath)$(NUnitExecutable) /nologo $(TestAssemblies) /exclude:IntegrationTest /xml=$(NUnitTestResultXmlFiles) /framework=$(TargetFrameworkVersion)" Condition=" '$(TEAMCITY_VERSION)' == '' and '$(OS)' == 'Windows_NT' " />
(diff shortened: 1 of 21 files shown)


fir3pho3nixx:
This *might* close #174

buildscripts/Build.proj
```diff
@@ -207,24 +221,37 @@ limitations under the License.
 
 	</Target>
 
-
 	<Target
-		Name="_ExecStatLight"
+		Name="_ExecNUnit_Without_IntegrationTest"
```
jonorossi: It would be great if we didn't have a copy of the `_ExecNUnit` target just to change the excluded categories, maybe just set a property for `AdditionalNUnitArguments` and call `_ExecNUnit`. However, I'm happy enough to merge this as is.

buildscripts/Build.proj
```diff
@@ -207,24 +221,37 @@ limitations under the License.
...
+		<CreateProperty Value="v4.0" Condition="'$(TargetFrameworkVersion)' == 'v4.5'">
+			<Output PropertyName="TargetFrameworkVersion" TaskParameter="Value" />
 		</CreateProperty>
+		
+		<MakeDir Directories="$(TestResultsPath)" Condition="'$(TestResultsPath)' != '' And !Exists('$(TestResultsPath)')" />
 
-		<Exec Command="$(StatLightCmdLine)" />
+		<Message Text="Running tests from assemblies: $(TestAssemblies)" />
+		
+		<Exec Command="$(NUnitPath)$(NUnitExecutable) /nologo $(TestAssemblies) /exclude:IntegrationTest,Explicit /xml=$(NUnitTestResultXmlFiles) /framework=$(TargetFrameworkVersion)" Condition=" '$(TEAMCITY_VERSION)' == '' and '$(OS)' == 'Windows_NT' " />
```
jonorossi: Where does the `Explicit` category get used? Tests marked with `Explicit` are already handled by the NUnit runner.

buildscripts/Build.proj
fir3pho3nixx: Glad to hear it! This was just to get it going. I have a feeling we would need to refine this anyway once we start embedding dotnet core elements. So we will be revisiting this.

buildscripts/Build.proj
fir3pho3nixx: You are right, took this out.

jonorossi:
Great, merged.

jonorossi:
At least the all tests side works well 👍 , unit test count for .NET 4.5 stayed at 1389:
http://builds.castleproject.org/viewType.html?buildTypeId=bt132

Issue #165, opened 2017-02-06
QUESTION: What is the future of this project?


fir3pho3nixx:
Hi, 

To start, I absolutely swear by this project. It has been my go to place for IoC. I absolutely hate all other containers as a result. They suck for many reasons. This one totally does not. That said please take the following with a pinch of salt.  

I would like to propose a `redux` but before that can happen I need to know where you guys are heading. There is a moving target when it comes to dotnet core, it is hard to anticipate. I can also see someone has had a go at porting the project which is great, although it might be a bit premature to try (happy to be told off for thinking this). 

After forking the project, I noticed a couple of things: 

 - WCF is a thing. Does it have to be? Can we delete this facility? It adds weight to contributions in terms of ramp up and does not make sense given the momentum in dotnet core. WCF is playing catch up and let's face it nobody cares about it anymore. IoC containers should not be marred by things like running `netsh.exe` and enabling port sharing using `sc.exe config NetTcpPortSharing` to make tests pass. Can we drop this and get back to basics in anticipation of developments in core? With the advent of DevOps and PaaS we *honestly* don't give a monkeys undies.

 - Silverlight is a thing. Does it really, really have to be? I mean really? RIA is deadskies. Can we remove this pretty please! :)

 - Prescription of logging is a thing. In dotnet core there might be a revolution when it comes to logging.  In cloud there already is. In Azure we delete most logging frameworks and implement Application Insights(which I must say is the one thing Microsoft did right). In AWS we use CloudWatch. There are also a multitude of log shipping vendors based on this. Why oh why are we tied down to log4net and NLog? If we remove this and substitute it with extensible lambda based extentions or leverage sub dependency resolvers better, surely we can get rid of this and anticipate core better? Docs become lighter ... 

 - Tests are failing in master(apart from all the crap to do with WCF). FluentRegistrationTestCase.Can_publish_events_via_AllTypes, would fix it if I new where we are going. AllTypes has been marked as obsolete. So should we delete this and start thinking about getting rid of obsolescence?  

 - Test have been ignored or are made explicit. Is this a change of direction? Specifically scope resolution and convention validation. Wwwwwwwhhhhhhhhyyyyyyyyyyy? Master should be kept pristine. Always!

So, now that you hate me, how do I help? 

Gav

jonorossi:
[...]

PR #180, opened 2017-02-20
Final clean up of silverlight

fir3pho3nixx:
@jonorossi 

Here are the final things I found laying around to do with silverlight(#165). 

When I grepWin the root folder for 'silverlight', the only things that popup are .cs files.


fir3pho3nixx:
Okiday, think we are good with the silverlight thing! 

Will dip in and out as my day job gets in the way ...

jonorossi:
Thanks @fir3pho3nixx, I had a feeling there was a lot more you'd missed, knew we had Silverlight stuff all over the place.

PR #183, opened 2017-02-21
Moved Castle.Core compatibility layer down to Castle.Windsor

fir3pho3nixx:
Moved all compatibility classes from Castle.Core down to Windsor for anticipation of migration to dotnet core.

### DIFF
--- a/src/Castle.Windsor/Compatibility/CustomAttributeExtensions.cs
+++ b/src/Castle.Windsor/Compatibility/CustomAttributeExtensions.cs
@@ -1,0 +1,58 @@
+﻿// Copyright 2004-2017 Castle Project - http://www.castleproject.org/
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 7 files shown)


src/Castle.Windsor/Compatibility/CustomAttributeExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#if (MONO || DOTNET35 || DOTNET40 || DOTNET45 || CLIENTPROFILE)
+
+namespace System.Reflection
+{
+	using System.Collections.Generic;
+
+	// This allows us to use the new reflection API while still supporting .NET 3.5 and 4.0.
```
jonorossi: I think it is about time we discussed the plans for Windsor v4, we are dropping all Silverlight and I think we should drop .NET Framework 3.5 and 4.0 including any client profile ugliness... see #185.

That being that case, much of this isn't needed because .NET 4.5 added `GetTypeInfo` and some friends. You can even see that the .NET 4.5 build of Castle Core doesn't define `FEATURE_LEGACY_REFLECTION_API` (https://github.com/castleproject/Core#conditional-compilation-symbols).

PR #197, opened 2017-03-06
Removed DOTNET40 build flags from code

fir3pho3nixx:
This one was luckily very small :)


src/Castle.Facilities.WcfIntegration.Tests/Service/Discovery/LoadBalance/FirstAvailablePolicyTestCase.cs
```diff
@@ -1,107 +0,0 @@
...
-// 
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-namespace Castle.Facilities.WcfIntegration.Tests
-{
-#if DOTNET40
```
jonorossi: These tests probably should stay. The `DOTNET40` symbol used to be defined for both .NET 4.0 and .NET 4.5 builds, so these went into the .NET 4.5 builds.

fir3pho3nixx: Oops, my bad. Got a bit too happy with the delete key. Will sort this.

PR #199, opened 2017-03-06
Removed MONO build flags from code (Merge this last)

fir3pho3nixx:
Okiday, all clean

### DIFF
--- a/src/Castle.Windsor/MicroKernel/SubSystems/Conversion/ComponentModelConverter.cs
+++ b/src/Castle.Windsor/MicroKernel/SubSystems/Conversion/ComponentModelConverter.cs
@@ -29,12 +29,6 @@ namespace Castle.MicroKernel.SubSystems.Conversion
 	{
 		public override bool CanHandleType(Type type)
 		{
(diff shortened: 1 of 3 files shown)


src/Castle.Windsor/MicroKernel/SubSystems/Conversion/ComponentModelConverter.cs
```diff
@@ -29,12 +29,6 @@ public class ComponentModelConverter : AbstractTypeConverter
 	{
 		public override bool CanHandleType(Type type)
 		{
-			// Mono 1.9+ thinks it can convert strings to interface
-			if (type.IsInterface)
```
jonorossi: This test doesn't seem like a Mono only check. I assume you can't have a converter for an interface so doesn't this just speed up this method without the need to call `TypeDescriptor.GetConverter` which will always return null for interfaces?

fir3pho3nixx: Hmmm, sounds like we are missing a test here ... let me give it another look tonight.

PR #201, opened 2017-03-09
Removing DOTNET40 build flags but keeping tests

fir3pho3nixx:
Rather than commit reverts I did this again. You were right, getting rid of the tests was a misstep.


src/Castle.Facilities.WcfIntegration/Service/Discovery/AdHocServiceCatalogDiscovery.cs
```diff
@@ -1,139 +0,0 @@
...
-{
-#if DOTNET40
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.ServiceModel;
-using System.ServiceModel.Description;
-using System.ServiceModel.Discovery;
-
-	public class AdHocServiceCatalogProbe : AbstractServiceHostAware, IDisposable
```
jonorossi: Doesn't the same apply to this class? i.e. DOTNET40 was defined in both the .NET 4.0 and 4.5 configurations, so this was in .NET 4.5.

fir3pho3nixx: This little guy was not in the project file. It is was inspired me to scan for files not in the solution

jonorossi: Ah yes.

PR #203, opened 2017-03-09
Got rid of stuff that is not in the solution

fir3pho3nixx:
Let's do the right thing, and lighten the load ...


fir3pho3nixx:
I totally don't expect you to merge this, but I hope you will at least consider getting rid of "dead code" ... :)

jonorossi:
Yep, not going to merge, however I've created #205 to handle those files.

PR #206, opened 2017-03-10
Removed superfluous references from nuspec file for Castle Windsor

fir3pho3nixx:
Cleaned up nuspec files. 

I must say I am hesitant to remove any of the framework assemblies on the 'Castle.WcfIntegrationFacility.nuspec' file, purely because I *think* if I remove some of these things might break. So will leave this one for now if that is OK...


jonorossi:
Thanks.

RE: `Castle.WcfIntegrationFacility.nuspec`, it just has two copies of the exact same framework assemblies, one set for .NET 4.0 and another for 4.5.

PR #216, opened 2017-03-15
Fix the link XML configuration reference

CFMITL:
I checked the other 3 references to this file and they are all in lowercase correctly.


jonorossi:
Thanks @CFMITL, merged.

Issue #209, opened 2017-03-14
Resolve after ConfigureAwait(false) in ASP.NET application


lukas-lansky:
We are slowly moving a large codebase towards dependency injection. We decided that to be able to start creating services that can be used across the codebase, we will allow ourselves to do service-locatorish approach towards resolution in static contexts: we ask static composition root to ``Resolve<T>()`` our service. This is of course not optimal, but is allow us to move gradually. When static contexts fade away, we will forbid ourselves to do this.

Now, we are also started to introduce async-await and we are consistently applying ``ConfigureAwait(false)`` to be able to utilize the thread pool better. We know that we are going to loose better parts of ``HttpContext.Current`` and we don't care too much as we don't want to depend on it anyway. But Windsor's ``PerWebRequestLifestyleModule`` depends on it, so we are unable to resolve components after the first await transition.

When looking into Windsor source code, I see that you use ``CallContext`` for saving state at ``CallContextScope`` and I wonder whether you thought about saving ``PerWebRequest`` state into it instead of ``HttpContext.Current.Items``. Isn't this something that ASP.NET Core support is going to demand anyway as the role of ``HttpContext`` is diminshed there? (Not sure at all, honest question.)

jonorossi:
I don't remember exactly, but I think when @kkozmic did the refactor for scoped lifestyles for Windsor 3.0 the use of `CallContext` was added, the `PerWebRequest` lifestyle implementation definitely predates that and has always depended on `System.Web`. There is a bit more info in the docs: https://github.com/castleproject/Windsor/blob/master/docs/lifestyles.md.

With the history aside, yes the implementation would have to change for the `PerWebRequest` lifestyle to be used with ASP.NET Core, however changing it would likely be more work and than just using a simple scoped lifestyle, others have already made some progress with a discussion in #120.

jonorossi:
Obviously a lot has happened in the last few years with ASP.NET Core and Windsor's per web request lifestyle. Closing this as stale, feel free to reopen if there is new info.

Issue #103, opened 2015-10-14
Threading issue when using LifestyleScoped


robertturner:
I am getting an internal exception in when resolving using LifestyleScoped in a multi-threaded app.
I'm resolving instances via factories which call a scoped accessor. The exception doesn't occur in the accessor but when attempting to set the Burden. I get the following ArgumentException: "An item with the same key has already been added." with stack trace:

```
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Castle.MicroKernel.Lifestyle.Scoped.ScopeCache.set_Item(Object id, Burden value)
   at Castle.MicroKernel.Lifestyle.Scoped.DefaultLifetimeScope.GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
   at Castle.MicroKernel.Lifestyle.ScopedLifestyleManager.Resolve(CreationContext context, IReleasePolicy releasePolicy)
   at Castle.MicroKernel.Handlers.DefaultHandler.ResolveCore(CreationContext context, Boolean requiresDecommission, Boolean instanceRequired, Burden& burden)
   at Castle.MicroKernel.Handlers.DefaultHandler.Resolve(CreationContext context, Boolean instanceRequired)
   at Castle.MicroKernel.Handlers.AbstractHandler.Resolve(CreationContext context)
   at Castle.MicroKernel.DefaultKernel.ResolveComponent(IHandler handler, Type service, IDictionary additionalArguments, IReleasePolicy policy)
   at Castle.MicroKernel.DefaultKernel.Castle.MicroKernel.IKernelInternal.Resolve(Type service, IDictionary arguments, IReleasePolicy policy)
   at Castle.Facilities.TypedFactory.TypedFactoryComponentResolver.Resolve(IKernelInternal kernel, IReleasePolicy scope)
   at Castle.Facilities.TypedFactory.Internal.TypedFactoryInterceptor.Resolve(IInvocation invocation)
   at Castle.Facilities.TypedFactory.Internal.TypedFactoryInterceptor.Intercept(IInvocation invocation)
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at Castle.Proxies.IPSECommonIOFactoryProxy.Create(IFtpObj ftpObj)
```

[...]

PR #221, opened 2017-04-19
Fix: An item with the same key has already been added.

hikalkan:
This may happen if createInstance 'somehow' recursively create the same instance. The exception I had:

````
System.ArgumentException: An item with the same key has already been added. Key: late bound IPropertyListener
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException(Object key)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Castle.MicroKernel.Lifestyle.Scoped.ScopeCache.set_Item(Object id, Burden value)
   at Castle.MicroKernel.Lifestyle.Scoped.DefaultLifetimeScope.GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
   at Castle.MicroKernel.Lifestyle.ScopedLifestyleManager.Resolve(CreationContext context, IReleasePolicy releasePolicy)
   at Castle.MicroKernel.Handlers.DefaultHandler.ResolveCore(CreationContext context, Boolean requiresDecommission, Boolean instanceRequired, Burden& burden)
   at Castle.MicroKernel.Handlers.DefaultHandler.Resolve(CreationContext context, Boolean instanceRequired)
   at Castle.MicroKernel.DefaultKernel.ResolveComponent(IHandler handler, Type service, IDictionary additionalArguments, IReleasePolicy policy)
   at Castle.MicroKernel.DefaultKernel.Castle.MicroKernel.IKernelInternal.ResolveAll(Type service, IDictionary arguments, IReleasePolicy policy)
   at Castle.MicroKernel.Resolvers.DefaultDependencyResolver.Resolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
   at Castle.MicroKernel.ComponentActivator.DefaultComponentActivator.CreateConstructorArguments(ConstructorCandidate constructor, CreationContext context)
   at Castle.MicroKernel.ComponentActivator.DefaultComponentActivator.Instantiate(CreationContext context)
   at Castle.MicroKernel.ComponentActivator.DefaultComponentActivator.InternalCreate(CreationContext context)
   at Castle.MicroKernel.ComponentActivator.AbstractComponentActivator.Create(CreationContext context, Burden burden)
   at Castle.MicroKernel.Lifestyle.AbstractLifestyleManager.CreateInstance(CreationContext context, Boolean trackedExternally)
   at Castle.MicroKernel.Lifestyle.ScopedLifestyleManager.<>c__DisplayClass3_0.<Resolve>b__0(Action`1 afterCreated)
   at Castle.MicroKernel.Lifestyle.Scoped.DefaultLifetimeScope.GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
   at Castle.MicroKernel.Lifestyle.ScopedLifestyleManager.Resolve(CreationContext context, IReleasePolicy releasePolicy)
   at Castle.MicroKernel.Handlers.DefaultHandler.ResolveCore(CreationContext context, Boolean requiresDecommission, Boolean instanceRequired, Burden& burden)
   at Castle.MicroKernel.Handlers.DefaultHandler.Resolve(CreationContext context, Boolean instanceRequired)
   at Castle.MicroKernel.Resolvers.DefaultDependencyResolver.ResolveFromKernelByType(CreationContext context, ComponentModel model, DependencyModel dependency)
   at Castle.MicroKernel.Resolvers.DefaultDependencyResolver.Resolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
   at Castle.MicroKernel.ComponentActivator.DefaultComponentActivator.CreateConstructorArguments(ConstructorCandidate constructor, CreationContext context)
   at Castle.MicroKernel.ComponentActivator.DefaultComponentActivator.Instantiate(CreationContext context)
   at Castle.MicroKernel.ComponentActivator.DefaultComponentActivator.InternalCreate(CreationContext context)
   at Castle.MicroKernel.ComponentActivator.AbstractComponentActivator.Create(CreationContext context, Burden burden)
   at Castle.MicroKernel.Lifestyle.AbstractLifestyleManager.CreateInstance(CreationContext context, Boolean trackedExternally)
   at Castle.MicroKernel.Lifestyle.ScopedLifestyleManager.<>c__DisplayClass3_0.<Resolve>b__0(Action`1 afterCreated)
   at Castle.MicroKernel.Lifestyle.Scoped.DefaultLifetimeScope.GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
   at Castle.MicroKernel.Lifestyle.ScopedLifestyleManager.Resolve(CreationContext context, IReleasePolicy releasePolicy)
   at Castle.MicroKernel.Handlers.DefaultHandler.ResolveCore(CreationContext context, Boolean requiresDecommission, Boolean instanceRequired, Burden& burden)
   at Castle.MicroKernel.Handlers.DefaultHandler.Resolve(CreationContext context, Boolean instanceRequired)
   at Castle.MicroKernel.DefaultKernel.ResolveComponent(IHandler handler, Type service, IDictionary additionalArguments, IReleasePolicy policy)
   at Castle.MicroKernel.DefaultKernel.Castle.MicroKernel.IKernelInternal.Resolve(Type service, IDictionary arguments, IReleasePolicy policy)
   at Castle.Windsor.MsDependencyInjection.ScopedWindsorServiceProvider.GetServiceInternal(Type serviceType, Boolean isOptional)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider)
   at Microsoft.EntityFrameworkCore.Infrastructure.AccessorExtensions.GetService[TService](IInfrastructure`1 accessor)
   at Microsoft.EntityFrameworkCore.Internal.LazyRef`1.get_Value()
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.UnbufferedEntityShaper`1.Shape(QueryContext queryContext, ValueBuffer valueBuffer)
   at Microsoft.EntityFrameworkCore.Query.AsyncQueryMethodProvider.<>c__DisplayClass3_0`1.<_ShapedQuery>b__0(ValueBuffer vb)
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncLinqOperatorProvider.SelectAsyncEnumerable`2.SelectAsyncEnumerator.<MoveNext>d__4.MoveNext()
````

### DIFF
--- a/src/Castle.Windsor/MicroKernel/Lifestyle/Scoped/DefaultLifetimeScope.cs
+++ b/src/Castle.Windsor/MicroKernel/Lifestyle/Scoped/DefaultLifetimeScope.cs
@@ -42,10 +42,17 @@ namespace Castle.MicroKernel.Lifestyle.Scoped
 		public Burden GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
 		{
 			var burden = scopeCache[model];
+			
 			if (burden == null)
 			{
-				scopeCache[model] = burden = createInstance(onAfterCreated);
+				burden = createInstance(onAfterCreated);
+
+				if (scopeCache[model] == null)
+				{
+				    scopeCache[model] = burden;
+				}
 			}
+		    
 			return burden;
 		}
 	}

### CODE REVIEW

**Comment by jonorossi** [comment]
@hikalkan do you know how your code triggered this? I'm concerned this is masking another problem as 2 burdens would get created for the same component. That might not be a problem, but a unit test would be great because it'll at least make sure this problem doesn't get reintroduced by a simple refactoring in the future.

**Comment by hikalkan** [comment]
I got the exception from EF Core usage. EF does very differently uses DI. It uses some factory methods which internally resolves some services from container which causes such a recursive call.
In ``createInstance`` call, it somehow requests same service again but and adding to the cache before the first call insert into the cache. I don't know how it does it since it's source code is so complicated. I implemented this as a workaround and I thought this code change should not cause any other problem.
Anyway, I decided to do not integrate EF to Windsor for now and for me no problem if you don't merge it :)

**Comment by jonorossi** [comment]
@hikalkan okay, thanks for the extra info. Did you happen to look at the `model` parameter of the 3 nested calls to `GetCachedInstance`?

The fact that the top of the call stack is an async operation may also indicate a threading problem.

**Comment by jonorossi** [comment]
It looks like we've got a similar report (#103) when the scoped lifestyle is used in a multithreaded app.

**Comment by hikalkan** [comment]
Haven't checked parameters since I didn't debug it, just re-compiled Windsor and used compiled dll in my application. It can be related to threading issue, I'm not sure. If so, what could be the solution?

**Comment by fir3pho3nixx** [comment]
CallContextLifetimeScope uses the following thread safe implementation: 

```csharp

public Burden GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
		{
			using (var token = @lock.ForReadingUpgradeable())
			{
				var burden = cache[model];
				if (burden == null)
				{
					token.Upgrade();

					burden = createInstance(delegate { });
					cache[model] = burden;
				}
				return burden;
			}
		}

```

Aren't we looking for something similar?

fir3pho3nixx:
DefaultLifetimeScope.cs was designed as an upgrade to web request and default scoped lifestyles. 

Please see commit from @kkozmic: https://github.com/castleproject/Windsor/commit/ccf06cef1e89a6afaf996b32afb60c069601e6e1

I think the idea here was if the lifestyle scope is deferred out to the calling context, the calling context would know better how to handle it's own thread safety. Clearly with EF here this is not happening. 

Should we make this thread safe and see if it fixes the problem @hikalkan, @jonorossi?

@hikalkan raised a PR here also: https://github.com/cryosharp/fortress/pull/38

I could take the hit in fortress first, if this fixes it, we should then back port the change here.

jonorossi:
> CallContextLifetimeScope uses the following thread safe implementation

Yep, as does `ThreadStaticLifetimeScope` (which is only used on Silverlight in place of `CallContextLifetimeScope` because Silverlight doesn't have the `CallContext` APIs). It is also worth noting that `CallContextLifetimeScope`'s `Dispose` method uses the same reader writer lock to access its `ScopeCache`. `CallContextLifetimeScope` was designed to be passed to other threads.

@kkozmic [explicitly called out in the documentation](https://github.com/castleproject/Windsor/blob/master/docs/implementing-custom-scope.md#the-scope-abstraction-ilifetimescope) that `DefaultLifetimeScope` is not thread safe and provided an example of implementing a scope that is. It was a conscience design decision.

I don't doubt adding the locking will fix the problem you are seeing, however that does change the design which appears to work well for scopes that don't need multi-threaded scope resolution including the non-Core ASP.NET PerWebRequest lifestyle where the same worker thread is used for the whole request processing.

#103 ran into this same problem because they also wrote a custom `IScopeAccessor` designed for multi-threaded use but are using a scope not designed for multi-threaded use.

I think the problem here is that `Castle.Windsor.MsDependencyInjection` is also using `DefaultLifetimeScope` in an application also performing multi-threaded scope resolution. I don't know how threading of ASP.NET Core apps works but it appears that Entity Framework doesn't have a problem performing multi-threaded work for a single web request and so the new lifestyle needs to accommodate that design.

jonorossi:
As suggested by @lukas-lansky (in #209) it is probably worthwhile reworking the PerWebRequest lifestyle to use `CallContext` rather than `HttpContext` as async is starting to work its way into even non-Core ASP.NET, so your `Castle.Windsor.MsDependencyInjection` should probably do the same.

`DefaultLifetimeScope` is still probably good for things like per-thread lifestyle.

hikalkan:
I don't think that EF uses multi-threading but let me test it (I will change ``DefaultLifetimeScope`` code to log thread ids and other parameters).

hikalkan:
Hi,

It seems you are right, this is probably because of multi-threading. I changed Castle.Windsor.MsDependencyInjection's code like that: https://github.com/volosoft/castle-windsor-ms-adapter/commit/50cacd63e8ffbc386c44eafb2506bbf9b459ad32
And it seems fixed the issue.

Thank you all for your valuable time.

PR #222, opened 2017-04-20
Add new solution and new projects for VS2017

alinapopa:


src/Castle.Windsor/Castle.Windsor-VS2017.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <AssemblyName>Castle.Windsor</AssemblyName>
+    <RootNamespace>Castle</RootNamespace>
+      <AssemblyOriginatorKeyFile>../../buildscripts/CastleKey.snk</AssemblyOriginatorKeyFile> 
+     <SignAssembly>true</SignAssembly> 
+    <PackageId>Castle.Windsor</PackageId>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Castle.Core" Version="3.3.0.0" />
```
jonorossi: It seems a little strange to include the extra ".0" when the nuget version doesn't have it.

src/Castle.Facilities.EventWiring/Castle.Facilities.EventWiring-VS2017.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <AssemblyName>Castle.Facilities.EventWiring</AssemblyName>
+    <RootNamespace>Castle.Facilities.EventWiring</RootNamespace>
+    <AssemblyOriginatorKeyFile>../../buildscripts/CastleKey.snk</AssemblyOriginatorKeyFile> 
+    <SignAssembly>true</SignAssembly> 
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);DOTNET45</DefineConstants>
```
jonorossi: I assume something the @fir3pho3nixx's common.props file will come later since DefineConstants is undefined until defined, or do I misunderstand how this works?

Also is there a reason all these projects except the main Windsor project define DOTNET45?

alinapopa: There was no reason - I missed it on the main Windsor project. I added it there too.
Regarding `$(DefineConstants)`, it is the recommended way to use `<DefineConstants>$(DefineConstants);MY_CUSTOM_CONSTANT</DefineConstants>` and it keeps the project simple. I noticed that if I remove `$(DefineConstants)`, the TRACE constant goes away. 
I assume the constant DOTNET45 will be removed in the next iterations, does not seem to be needed anymore.

alinapopa:
Addressed the requested changes

src/Castle.Facilities.EventWiring/Castle.Facilities.EventWiring-VS2017.csproj
fir3pho3nixx: Yes, Alina, this is a work in progress in castle core too. If there is anything you think I should be doing while I am up in core please let me know. It very quickly gets out of control which is why @jonorossi is calling it out.

fir3pho3nixx:
This is some good stuff.  @jonorossi I am relying on you for oversight. @alinapopa I am relying on you to tell if we can do it better(happy to make changes in core), like it so far! 👍 

Before we merge this, can we consider getting some tests going?

appveyor.yml
```diff
@@ -0,0 +1,25 @@
+pull_requests:
+  do_not_increment_build_number: true
```
jonorossi: @fir3pho3nixx we should set this in Castle Core too, pull request build numbers are throw away so might as well get a random string appended to the end of the version.

src/Castle.Facilities.EventWiring/Castle.Facilities.EventWiring-VS2017.csproj
jonorossi: If `TRACE` is automatically defined, I assume that means `DEBUG` is also automatically defined by the targets files for a "Debug" configuration build?

If so, @fir3pho3nixx we should simplify the Castle Core one and let the tools define those two.

jonorossi: Yes, we won't need `DOTNET45` after we are done with the feature symbols to support .NET Standard, but obviously it is needed until we get that sorted after we get this build refactor merged to master.

PR #228, opened 2017-04-28
Add common.props and artifacts

alinapopa:
Add common.props
Add nuget package to artifacts
Upload test result xml's to artifacts

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -44,14 +44,3 @@ build_script:
 artifacts:
  - path: Packages\*.nupkg
  
-#test deploy to myget
-deploy:
-  - provider: NuGet
-    artifact: /.*\.nupkg/
-    server: https://www.myget.org/F/windsor_test/api/v2/package
-    api_key: 
-       secure: NulQ1qClvThpFTs6XZfer2DG8Qe2TjK0bt1zhAuvRXe9opfwJIbAEMr1ceUl/N79
-    skip_symbols: true
-    prerelease: true
-    on:
-     branch: refactor-build
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,4 @@
+
 pull_requests:
   do_not_increment_build_number: true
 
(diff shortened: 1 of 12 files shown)


common.props
```diff
@@ -0,0 +1,38 @@
...
+    <RepositoryUrl>https://github.com/castleproject/Windsor</RepositoryUrl>
+    <PackageLicenseUrl>http://www.apache.org/licenses/LICENSE-2.0.html</PackageLicenseUrl>
+    <PackageProjectUrl>http://www.castleproject.org/projects/windsor/</PackageProjectUrl>
+    <PackageIconUrl>http://www.castleproject.org/img/windsor-logo.png</PackageIconUrl>
+    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
+    <PackageOutputPath>../../Packages/</PackageOutputPath>
+    <PackageVersion>$(BuildVersion)</PackageVersion>
+    <IncludeSymbols>true</IncludeSymbols>
+    <IncludeSource>false</IncludeSource>
+    <Copyright>Copyright 2004 Castle Project - http://www.castleproject.org/ </Copyright>
```
jonorossi: This should be copyright 2004-{current_year}.

appveyor.yml
```diff
@@ -22,16 +23,35 @@ test: off
...
+    Push-AppveyorArtifact $env:APPVEYOR_BUILD_FOLDER\src\Castle.Facilities.WcfIntegration.Tests\$testBinPath\TestResult_WcfIntegration.xml
 
 build_script:  
   - cmd: build-vs2017.cmd
+
+artifacts:
+ - path: Packages\*.nupkg
+ 
+#test deploy to myget
+deploy:
```
jonorossi: I assume this was accidentally committed to this pull request.

common.props
```diff
@@ -0,0 +1,38 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);DOTNET45</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="..\..\BreakingChanges.txt" />
```
jonorossi: I see the packages contain a "content" and a "contentFiles" folder both with this set of files. How does this work, what do these directories do because we used to have these files in the root of the package?

If VS doesn't automatically open readme.txt we can get rid of that one.

alinapopa: I need to specify the `ContentTargetFolders` property or a `PackagePath` for the content files (by default these files end up in both "content" and "contentFiles"). If I set the `ContentTargetFolders` to be the root of the package, there is an issue with the LICENSE file, because it doesn't have an extension. Trying to open the resulting package with NuGet package Explorer gives the error: "Part URI is not valid per rules defined in the Open Packaging Convention specification". 
So there are two options:
1.Put these files in a folder named "content" or "contentFiles" or whatever name
2.Rename LICENSE to LICENSE.txt so that these files can be put in the root folder.

common.props
alinapopa: Added current year in common.props.

appveyor.yml
alinapopa: I removed it.

common.props
alinapopa: For now I added a PreBuildEvent to copy this file with the txt extension.

jonorossi: The two directories seems a bit hacky, but thanks for getting it sorted.

PR #229, opened 2017-05-04
Rename LICENSE to LICENSE.txt and get rid of the copying.

alinapopa:
The LICENSE file without an extension causes problems when packing it into NuGet package in the root folder. Renaming it so that copying it over with an extension is not necessary.


jonorossi:
It sort of sucks that NuGet has a problem with this, but oh well. Merged, thanks.

PR #231, opened 2017-05-08
Update tests to use NUnit3

alinapopa:
Update to NUnit3
Replace the ExpectedExceptionAttribute with Assert.Throws()


alinapopa:
Updating to NUnit3 is needed in order to use the latest adapter
http://www.alteridem.net/2017/05/04/test-net-core-nunit-vs2017/
I would also like to add the new test adapter so that I can use dotnet test for running the tests and also the tests will show up in Visual Studio.
https://www.nuget.org/packages/NUnit3TestAdapter/3.8.0-alpha1
@jonorossi are you ok with using the new test adapter now?

src/Castle.Windsor.Tests/Config/ConfigXmlInterpreterTestCase.cs
```diff
@@ -64,13 +64,12 @@ public void CorrectConfigurationMapping()
...
 
 		[Test]
-		[ExpectedException(typeof(ConfigurationProcessingException))]
 		public void MissingManifestResourceConfiguration()
 		{
 			var store = new DefaultConfigurationStore();
 			var source = new AssemblyResource("assembly://Castle.Windsor.Tests/missing_config.xml");
 			IKernel kernel = new DefaultKernel();
-			new XmlInterpreter(source).ProcessResource(source, store, kernel);
+            Assert.Throws<ConfigurationProcessingException>(() => new XmlInterpreter(source).ProcessResource(source, store, kernel));
```
jonorossi: The indention is off.

src/Castle.Windsor.Tests/Configuration2/Properties/PropertiesTestCase.cs
```diff
@@ -38,10 +38,10 @@ public void CorrectEval()
 		}
 
 		[Test]
-		[ExpectedException(typeof(ConfigurationProcessingException))]
 		public void MissingProperties()
 		{
-			container = new WindsorContainer(ConfigHelper.ResolveConfigPath("Configuration2/Properties/config_with_missing_properties.xml"));
+            Assert.Throws<ConfigurationProcessingException>(() =>
```
jonorossi: The indention is off.

src/Castle.Windsor.Tests/TypedParametersTestCase.cs
```diff
@@ -105,15 +105,18 @@ public void Typed_arguments_work_for_DynamicParameters_mixed()
...
+			Assert.Throws<ConverterException>(() =>
 			{
-				{ typeof(int), "not an int" },
-				{ typeof(string), "a string" }
-			}));
+                Kernel.Resolve<ClassWithArguments>(new Arguments(new Dictionary<Type, object>
+                {
+                    { typeof(int), "not an int" },
+                    { typeof(string), "a string" }
+                }));
```
jonorossi: The indentation is off.

jonorossi:
If you can fix the indentation in a few places everything looks great. Apologies for the delay.

> https://www.nuget.org/packages/NUnit3TestAdapter/3.8.0-alpha1
> are you ok with using the new test adapter now?

Yep, happy for you to upgrade it.

hammett:
Are we using .editorconfig? It could take care of these minor editor issues, for all PRs

alinapopa:
The indentation issues appeared because the source code uses tabs and I edited them in Visual Studio which used spaces. It would be nice if editorconfig could help with that.

alinapopa:
Switching to using the new adapter did not work well.
https://ci.appveyor.com/project/alinapopa/windsor/build/1.0.102
Adding the new adapter caused 20+ test failures, most of them with the error: 

>  Could not load file or assembly Microsoft.VisualStudio.TestPlatform.ObjectModel, Version=11.0.0.0 

. trx output is successfully uploaded, but looks like AppVeyor cannot read it. Nothing is displayed in Tests tab. 

So I will not add the new test adapter to this PR.

jeremymeng:
> Could not load file or assembly Microsoft.VisualStudio.TestPlatform.ObjectModel, Version=11.0.0.0

After adding the new NUnit test adapter, this error is caused by the new `NUnit3.TestAdapter.dll` that are now dropped to the output folder.  This dll has a dependency on Microsoft.VisualStudio.TestPlatform.ObjectModel.  The failing tests scan all assemblies in the test output folder and use Refelection to get types from them.

There are several options

1. Don't add the test adapter. We could use NUnitLite in the same way Castle.Core does. We will be missing the Visual Studio Test Explorer integration though.
2. Change `GetAvailableTypes()` to catch and ignore exceptions when assemblies fail to load. [Current code](https://github.com/castleproject/Windsor/blob/9f07453e2ac27835fe4561599654575007b44c0e/src/Castle.Windsor/Core/Internal/ReflectionUtil.cs#L174) already catches `RefelectionTypeLoadException` and returns types loaded so far. This method is in `Castle.Core.Internal` so not a breaking change to consumers.
3. Change those tests to filter `NUnit3.TestAdapter.dll`. This would make tests ugly...
4. Copy assemblies to be scanned to a new folder and test that folder. Extra code to workaround this issue.

I think option 2 is reasonable.  @jonorossi any more ideas?

jonorossi:
> Are we using .editorconfig? It could take care of these minor editor issues, for all PRs

Hmmm, actually no. I assumed we did, but we've only got one in Castle Core.

jonorossi:
> After adding the new NUnit test adapter, this error is caused by the new NUnit3.TestAdapter.dll that are now dropped to the output folder. This dll has a dependency on Microsoft.VisualStudio.TestPlatform.ObjectModel. The failing tests scan all assemblies in the test output folder and use Refelection to get types from them.

After a quick look it appears `GetExportedTypes` gets the public types of the assembly that contains the `IWindsorInstaller`, in this case in the unit test assembly which must cause `NUnit3.TestAdapter` to get loaded.

I assume the `NUnit3TestAdapter` package is designed for Visual Studio, does it then only break running in VS or does this ship with the command line build too?

alinapopa:
@jonorossi It also breaks when running from command line, with `dotnet test`

alinapopa:
#234

PR #232, opened 2017-05-10
Add CLSCompliant assembly attribute

alinapopa:
Add CLSCompliant assembly attribute
Set build script return value


jonorossi:
> Until now I've tried to not make changes that would break old solution/build setup.

Since @fir3pho3nixx was able to sort out Mono with the new tools for Castle Core we've decided to dump all the old build set up, so we can do the same thing here too, should really make things easier.

Let's add the attribute to `buildscripts/CommonAssemblyInfo.cs`, and just make that file what we need for the new tooling.

alinapopa:
> Let's add the attribute to buildscripts/CommonAssemblyInfo.cs, and just make that file what we need for the new tooling.

@jonorossi We don't use the file `buildscripts\CommonAssemblyInfo.cs` in the new build setup (so it won't be combining two files in one). I can replace `AssemblyAttributes.cs` with `CommonAssemblyInfo.cs`.

PR #234, opened 2017-05-16
Use the latest NUnitTestAdapter for running tests

alinapopa:
Add NUnit3TestAdapter
Use `dotnet test` for running the tests
Upload test results in trx files instead of xml
Fix the error `Could not load file or assembly Microsoft.VisualStudio.TestPlatform.ObjectModel, Version=11.0.0.0` by catching FileNotFoundException in `GetAvailableTypes()`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,5 @@ _ReSharper*
 *.sln.DotSettings.user
 *.DotSettings.user
 /Packages
+/src/*/TestResults
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,15 +28,13 @@ on_finish:
 - ps: >-
     $wc = New-Object System.Net.WebClient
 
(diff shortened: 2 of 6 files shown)


fir3pho3nixx:
The only concern I have with doing this is that approach is not yet compatible with linux. 

Please see issue I raised here: https://github.com/nunit/nunit3-vs-adapter/issues/324

fir3pho3nixx:
@alinapopa - Can we close this out and track it with an issue? If we don't do this I can bring TravisCI to Windsor in `refactor-build` with netcore tests running both on windows and linux.

If you agree. Please create a new issue for this and drop it into the planned column and assign me to it. The same problem exists for Castle Core.

alinapopa:
Closing for now; opened issue #243

fir3pho3nixx:
Cool, added NUnit issue tracking link. Thanks.

PR #242, opened 2017-05-18
Rename LICENSE.txt to LICENSE

alinapopa:


alinapopa:
@fir3pho3nixx 
> What about the LICENSE.txt file?

Was renamed to LICENSE (no extension).

fir3pho3nixx:
I see. My bad :)

PR #244, opened 2017-05-19
Use GetTypeInfo() for Type methods in Castle.Windsor project.

alinapopa:
Use GetTypeInfo() for Type methods

### DIFF
--- a/src/Castle.Windsor/Core/CastleComponentAttribute.cs
+++ b/src/Castle.Windsor/Core/CastleComponentAttribute.cs
@@ -15,6 +15,7 @@
 namespace Castle.Core
 {
 	using System;
+	using System.Reflection;
 
 	/// <summary>
 	///   This attribute is useful only when you want to register all components
@@ -58,7 +59,7 @@ namespace Castle.Core
 
 		public static CastleComponentAttribute GetDefaultsFor(Type type)
 		{
-			var attribute = (CastleComponentAttribute)GetCustomAttribute(type, typeof(CastleComponentAttribute));
+			var attribute = (CastleComponentAttribute)type.GetTypeInfo().GetCustomAttribute(typeof(CastleComponentAttribute));
 			if (attribute != null)
 			{
 				if (attribute.ServicesSpecifiedExplicitly == false)
--- a/src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs
+++ b/src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs
@@ -18,6 +18,7 @@ namespace Castle.MicroKernel.Registration
 	using System.Collections.Generic;
 	using System.ComponentModel;
 	using System.Linq;
+	using System.Reflection;
 
 	using Castle.Core;
 	using Castle.MicroKernel.Lifestyle.Scoped;
@@ -456,13 +457,13 @@ namespace Castle.MicroKernel.Registration
 			var actuallyBasedOn = new List<Type>();
 			foreach (var potentialBase in potentialBases)
 			{
-				if (potentialBase.IsAssignableFrom(type))
+				if (potentialBase.GetTypeInfo().IsAssignableFrom(type))
 				{
 					actuallyBasedOn.Add(potentialBase);
 				}
-				else if (potentialBase.IsGenericTypeDefinition)
+				else if (potentialBase.GetTypeInfo().IsGenericTypeDefinition)
 				{
-					if (potentialBase.IsInterface)
+					if (potentialBase.GetTypeInfo().IsInterface)
 					{
 						if (IsBasedOnGenericInterface(type, potentialBase, out baseTypes))
 						{
@@ -517,14 +518,14 @@ namespace Castle.MicroKernel.Registration
 		{
 			while (type != null)
 			{
-				if (type.IsGenericType &&
+				if (type.GetTypeInfo().IsGenericType &&
 				    type.GetGenericTypeDefinition() == basedOn)
 				{
 					baseTypes = new[] { type };
 					return true;
 				}
 
-				type = type.BaseType;
+				type = type.GetTypeInfo().BaseType;
 			}
 			baseTypes = null;
 			return false;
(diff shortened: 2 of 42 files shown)


src/Castle.Windsor/Core/CastleComponentAttribute.cs
```diff
@@ -58,7 +59,7 @@ public bool HasName
 
 		public static CastleComponentAttribute GetDefaultsFor(Type type)
 		{
-			var attribute = (CastleComponentAttribute)GetCustomAttribute(type, typeof(CastleComponentAttribute));
+			var attribute = type.GetTypeInfo().GetCustomAttribute(typeof(CastleComponentAttribute)) as CastleComponentAttribute;
 			if (attribute != null)
```
jeremymeng: keep the direct cast

src/Castle.Windsor/MicroKernel/SubSystems/Conversion/AttributeAwareConverter.cs
```diff
@@ -67,7 +68,7 @@ private ITypeConverter TryGetConverterInstance(Type type)
 		{
 			ITypeConverter converter = null;
 
-			var attr = (ConvertibleAttribute)Attribute.GetCustomAttribute(type, typeof(ConvertibleAttribute));
+			var attr = type.GetTypeInfo().GetCustomAttribute(typeof(ConvertibleAttribute)) as ConvertibleAttribute;
```
jeremymeng: keep the direct cast

src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs
```diff
@@ -535,11 +536,10 @@ private static bool IsBasedOnGenericInterface(Type type, Type basedOn, out Type[
...
 			foreach (var @interface in type.GetInterfaces())
 			{
-				if (@interface.IsGenericType &&
+				if (@interface.GetTypeInfo().IsGenericType &&
 				    @interface.GetGenericTypeDefinition() == basedOn)
 				{
-					if (@interface.ReflectedType == null &&
-					    @interface.ContainsGenericParameters)
+					if (@interface.GetTypeInfo().ContainsGenericParameters)
 					{
```
jeremymeng: Should we wrap ReflectedType in a #if? It is in net45, and also added in netstandard2.0.

src/Castle.Windsor/MicroKernel/Registration/ServiceDescriptor.cs
```diff
@@ -229,14 +230,14 @@ private IEnumerable<Type> GetTopLevelInterfaces(Type type)
...
-			if (!serviceType.IsInterface)
+			if (!serviceType.GetTypeInfo().IsInterface)
 			{
 				return serviceType;
 			}
 			// This is a workaround for a CLR bug in
 			// which GetInterfaces() returns interfaces
 			// with no implementations.
-			if (serviceType.IsGenericType && serviceType.ReflectedType == null)
+			if (serviceType.GetTypeInfo().IsGenericType)
```
jeremymeng: Should we wrap ReflectedType in a #if? It is in net45, and also added in netstandard2.0.

src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs
alinapopa: Added #if FEATURE_REFLECTION_REFLECTEDTYPE

src/Castle.Windsor/Core/CastleComponentAttribute.cs
alinapopa: Fixed

src/Castle.Windsor/MicroKernel/SubSystems/Conversion/AttributeAwareConverter.cs
alinapopa: Fixed

src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs
alinapopa: Added #if FEATURE_REFLECTION_REFLECTEDTYPE

src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs
```diff
@@ -535,11 +536,14 @@ private static bool IsBasedOnGenericInterface(Type type, Type basedOn, out Type[
...
-				if (@interface.IsGenericType &&
+				if (@interface.GetTypeInfo().IsGenericType &&
 				    @interface.GetGenericTypeDefinition() == basedOn)
 				{
-					if (@interface.ReflectedType == null &&
-					    @interface.ContainsGenericParameters)
+					if (@interface.GetTypeInfo().ContainsGenericParameters
+#if FEATURE_REFLECTION_REFLECTEDTYPE
+						&& @interface.ReflectedType == null
+#endif
```
jonorossi: I'm not a fan of the two usages of `FEATURE_REFLECTION_REFLECTEDTYPE`, this isn't conditionally compiling functionality out (e.g. serialization or System.Web support) but something that shouldn't really differ between the builds on both runtimes otherwise it could introduce subtle bugs.

I don't know if DeclaringType could be used in place of ReflectedType here, but we removed the one usage of ReflectedType from Castle Core in https://github.com/castleproject/Core/commit/87c36c31490be1a7638ce4b0fb5441b5c4854c89 by passing the object's type around.

alinapopa: Because in both these places `ReflectedType` is just used to check for null (and not actually doing anything with the ReflectedType), I would say it doesn't make a difference if it's the base or derived type returned and therefore replacing it with `DeclaringType` is safe.

jonorossi: Sounds good to me. I think the only scenario where `ReflectedType` would return null would be for global methods on a module, and `DeclaringType` would also be null.

alinapopa: Replaced ReflectedType with DeclaringType

PR #246, opened 2017-05-19
Remove ICloneable

alinapopa:
Remove ICloneable

### DIFF
--- a/src/Castle.Windsor.Tests/MicroKernel/ArgumentsTestCase.cs
+++ b/src/Castle.Windsor.Tests/MicroKernel/ArgumentsTestCase.cs
@@ -102,7 +102,6 @@ namespace Castle.Windsor.Tests.MicroKernel
 			Assert.AreEqual(value, arguments["boo!"]);
 		}
 
-#if !SILVERLIGHT
 		[Test]
 		public void Custom_stores_get_picked_over_default_ones_in_clone()
 		{
@@ -112,11 +111,10 @@ namespace Castle.Windsor.Tests.MicroKernel
 
 			arguments.Add(key, value);
 
-			var clone = (IDictionary)((ICloneable)arguments).Clone();
+			var clone = arguments.Clone();
 
 			Assert.AreEqual(value, clone["boo!"]);
 		}
-#endif
 
 		[Test]
 		public void Handles_Type_as_key()
--- a/src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs
+++ b/src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs
@@ -364,13 +364,14 @@ namespace Castle.MicroKernel.Handlers
 			var extendedProperties = ComponentModel.ExtendedProperties;
 			if (extendedProperties != null && extendedProperties.Count > 0)
 			{
-#if !SILVERLIGHT
-				if (extendedProperties is ICloneable)
+				if (extendedProperties is Arguments)
 				{
-					extendedProperties = (IDictionary)((ICloneable)extendedProperties).Clone();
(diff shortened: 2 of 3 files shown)


src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs
```diff
@@ -364,7 +364,7 @@ private IDictionary GetExtendedProperties()
 			var extendedProperties = ComponentModel.ExtendedProperties;
 			if (extendedProperties != null && extendedProperties.Count > 0)
 			{
-#if !SILVERLIGHT
+#if FEATURE_ICLONEABLE
 				if (extendedProperties is ICloneable)
 				{
 					extendedProperties = (IDictionary)((ICloneable)extendedProperties).Clone();
```
jonorossi: I know I'm making you do more work than just a straight port 😄 , but I think it is valuable fixing some tech debt as .NET Core will be the primary target for new development so should work optimally.

As it is well known now, Microsoft discourages implementing `ICloneable` on a public API. We aren't really doing that here since it is explicitly implemented but we can get rid of the usage anyway. Interestingly as far as I can tell it isn't a deep clone just new internal dictionary that it wraps with the same items.

Since `ComponentModel.ExtendedProperties` is always an instance of `Arguments` (unless you are strangely creating your own `ComponentModel` instances) we can just do a check for that type instead of `ICloneable`

So keeping `Clone` without `ICloneable`:
- Remove `Arguments` implementing `ICloneable`
- Make `Arguments.Clone` internal, it is already an explicitly implemented anyway
- Fix this if-statement so it doesn't double create an `Arguments` class, it currently creates one when cloning then another immediately afterwards.

... after I wrote the last line I'm now looking at the code and really can't work out why it needs to be cloneable since the `Arguments` constructor takes a `IDictionary` and will just populate a new internal dictionary with everything, all the extra comparer stuff is confusing.

- Commit this was added: https://github.com/castleproject/Windsor/commit/c25a4f33eec9a1e73135827e1c3a83bc9b3f0ec2
- Old issue it makes reference to: https://github.com/castleproject/castle-youtrack-export/blob/master/IOC/IOC-282.xml

I'm actually now inclined to remove the cloning stuff, especially if that is the only unit test that fails, it is a unit tests for unit test sake.

alinapopa: @jonorossi I made now Clone() internal and kept it because otherwise that comparer will be lost and I don't know what is the effect of that.. Do you want to remove it all? So `GetExtendedProperties()` will just return `new Arguments(extendedProperties)`

src/Castle.Windsor.Tests/MicroKernel/ArgumentsTestCase.cs
```diff
@@ -102,22 +102,6 @@ public void Custom_stores_get_picked_over_default_ones()
...
-		[Test]
-		public void Custom_stores_get_picked_over_default_ones_in_clone()
-		{
-			var arguments = new Arguments(new CustomStringComparer());
-			var key = "foo";
-			var value = new object();
-
-			arguments.Add(key, value);
-
-			var clone = (IDictionary)((ICloneable)arguments).Clone();
```
jonorossi: I thought this unit test would still work, just remove the cast to `ICloneable`?

alinapopa: Yes but it needs `Clone()` to be public.

jonorossi: Ah yes, let's just make `Clone` public.

alinapopa: Done

PR #249, opened 2017-05-23
Add FEATURE_SECURITY_PERMISSIONS

alinapopa:
Add conditional compilation symbol FEATURE_SECURITY_PERMISSIONS

### DIFF
--- a/src/Castle.Windsor/MicroKernel/ComponentActivator/DefaultComponentActivator.cs
+++ b/src/Castle.Windsor/MicroKernel/ComponentActivator/DefaultComponentActivator.cs
@@ -18,7 +18,9 @@ namespace Castle.MicroKernel.ComponentActivator
 	using System.Reflection;
 	using System.Runtime.Serialization;
 	using System.Security;
+#if FEATURE_REMOTING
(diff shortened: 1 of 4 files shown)


src/Castle.Windsor/MicroKernel/ComponentActivator/DefaultComponentActivator.cs
```diff
@@ -173,7 +173,7 @@ protected object CreateInstanceCore(ConstructorCandidate constructor, object[] a
 			return instance;
 		}
 
-#if (!SILVERLIGHT)
+#if FEATURE_SECURITY_PERMISSIONS
```
jonorossi: Maybe it would be better to use `FEATURE_SERIALIZATION` or `FEATURE_REMOTING` since the core of why this can't work on .NET Standard is `FormatterServices.GetUninitializedObject` which is for binary serialization.

alinapopa: I replaced it with FEATURE_REMOTING.

PR #250, opened 2017-05-23
Harmonising Castle Windsor with Castle Core

fir3pho3nixx:
Harmonised the look and feel of the solution with Castle Core, great to have things looking the same. Also moved build output to .\build in keeping with old build conventions. common.props lives in buildscripts for core so moved this across also. Made sure the build versions also behave similarly to castle core, local builds always have 0.0.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,6 @@ _ReSharper*
 *.sln.DotSettings.user
 *.DotSettings.user
 /Packages
+buildscripts/**/obj/
+buildscripts/**/bin/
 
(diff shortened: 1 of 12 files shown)


jonorossi:
Looks good, I'll merge this now. @alinapopa if you've got any review comments they can still easily be made.

alinapopa:
Looks good. The only comment I have is to remove `Packages` from gitignore if there is no folder with this name now.

alinapopa:
Also what is buildscripts.csproj for?

fir3pho3nixx:
@alinapopa - will fix gitignore in my next PR, no problem. The buildscripts.csproj is a conventional project which `auto` includes everything inside the BuildScripts folder into the solution so you can conveniently edit things inside VS2017 without having to diving out to the file system the whole time. It also doubles as a NuGet installer for things that are required by the build but are not referenced anywhere, an example would be something like NUnit Console(albeit I have not done it here just yet).

PR #257, opened 2017-05-24
Add conditional compilation symbol FEATURE_SYSTEM_CONFIGURATION

alinapopa:


src/Castle.Facilities.Synchronize/CreateOnUIThreadInspector.cs
```diff
@@ -171,8 +173,11 @@ private static IReference<IProxyGenerationHook> ObtainProxyHook(IConfiguration c
 						var message = String.Format("The specified controlProxyHook does " +
 						                            "not implement the interface {1}. Type {0}",
 						                            hookType.FullName, typeof(IProxyGenerationHook).FullName);
-
+#if FEATURE_SYSTEM_CONFIGURATION
 						throw new ConfigurationErrorsException(message);
+#else
+						throw new InvalidOperationException(message);
+#endif
```
alinapopa: @jonorossi What would be a good exception to throw here when `ConfigurationErrorsException` is not available?

jonorossi: You could just change it to `System.Exception` in all cases because it doesn't quite sound like `ConfigurationErrorsException` should have been used. However it is all a bit moot because this whole file needs to be wrapped in `#if FEATURE_WINFORMS`.

alinapopa: I've wrapped the file in `#if FEATURE_WINFORMS`

src/Castle.Facilities.Synchronize/Castle.Facilities.Synchronize-VS2017.csproj
```diff
@@ -33,4 +33,8 @@
   <ItemGroup>
     <Compile Include="..\..\buildscripts\CommonAssemblyInfo.cs" />
   </ItemGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)'=='net45'">
+    <DefineConstants>$(DefineConstants);FEATURE_WINFORMS</DefineConstants>
```
jonorossi: I'm curious why you defined `FEATURE_WINFORMS` in here, we're very unlikely to ever want to build the .NET Framework builds without this code and it is surfaced more clearly if they are all in `common.props`.

alinapopa: It's only needed in project Castle.Facilities.Synchronize project and in tests. I'll move it to common.props if it's preferable that way.

alinapopa: Moved it to common.props

jonorossi: Thanks

PR #258, opened 2017-05-24
Replace Uri.SchemeDelimiter

alinapopa:


src/Castle.Windsor/MicroKernel/SubSystems/Configuration/DefaultConfigurationStore.cs
```diff
@@ -181,7 +181,8 @@ public IConfiguration[] GetInstallers()
 
 		public IResource GetResource(String resourceUri, IResource resource)
 		{
-			if (resourceUri.IndexOf(Uri.SchemeDelimiter) == -1)
+			var uriSchemeDelimiter = "://";
```
fir3pho3nixx: Instead of hardcoding these values, is this perhaps something we should think about putting in the compatibility layer? You never know, it might end up being a thing again in subsequent releases of netstandard.

alinapopa: Uri.SchemeDelimiter comes back in 2.0. @jonorossi  at you ok with adding something like  `FEATURE_URIMETHODS` for this?

jonorossi: Looks like it is defined in `System.Private.Uri`.

[`public static readonly string SchemeDelimiter = "://";`](https://github.com/dotnet/corefx/blob/7df6b74cc4aeebc86e76b5c7901a59fa7fa793f6/src/System.Private.Uri/src/System/Uri.cs#L29)

Since it is a static field you can't add it to the compatibility code as is, probably best to still define it as a const or readonly field though.

fir3pho3nixx: Cool cool... just I have been here before and was wondering if we wanted create a shim for it. If not all good...

jonorossi: > Uri.SchemeDelimiter comes back in 2.0. @jonorossi at you ok with adding something like FEATURE_URIMETHODS for this?

Hmm, GitHub didn't tell me you commented after I loaded the page. I guess I'm looking in the wrong place for .NET Standard 2.0 APIs.

No idea how long we'll support .NET Standard 1.x for. Go for it, would be good to get `Uri.SchemeDelimiter` explicitly mentioned in Windsor's list of conditional symbols like Castle Core.

alinapopa: https://apisof.net/catalog/System.Uri.SchemeDelimiter 
https://github.com/dotnet/corefx/issues/10215

alinapopa: Added conditional symbol

jonorossi: Thanks. It looks like GitHub search is broken because it doesn't return `System.Runtime.cs` even though the change is in master: https://github.com/dotnet/corefx/search?q=SchemeDelimiter

PR #260, opened 2017-05-25
Refactor build cleanup

fir3pho3nixx:
Cutting everything over after checking NuGet's


fir3pho3nixx:
This is weird, lost status checks

PR #262, opened 2017-05-25
Add FEATURE_SERIALIZATION

alinapopa:
Added FEATURE_SERIALIZATION.
I did not wrap `[Serializable]` attribute at this time because it builds on netstandard, when using [Serialization.Formatters](https://www.nuget.org/packages/System.Runtime.Serialization.Formatters/4.3.0) package.


jonorossi:
`System.Runtime.Serialization.Formatters` cool. Looks good.

PR #264, opened 2017-05-25
Replace Delegate.CreateDelegate with MethodInfo.CreateDelegate

alinapopa:
Replace `Delegate.CreateDelegate` with `MethodInfo.CreateDelegate`

### DIFF
--- a/src/Castle.Windsor/MicroKernel/ModelBuilder/Inspectors/LifestyleModelInspector.cs
+++ b/src/Castle.Windsor/MicroKernel/ModelBuilder/Inspectors/LifestyleModelInspector.cs
@@ -193,9 +193,9 @@ namespace Castle.MicroKernel.ModelBuilder.Inspectors
 						scopeRootBinderType.Name, name));
 			}
 			var instance = scopeRootBinderType.CreateInstance<object>();
-			return
-				(Func<IHandler[], IHandler>)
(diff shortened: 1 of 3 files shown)


src/Castle.Windsor/MicroKernel/ModelBuilder/Inspectors/LifestyleModelInspector.cs
```diff
@@ -193,9 +193,9 @@ protected virtual void ValidateTypeFromAttribute(Type typeFromAttribute, Type ex
 						scopeRootBinderType.Name, name));
 			}
 			var instance = scopeRootBinderType.CreateInstance<object>();
-			return
-				(Func<IHandler[], IHandler>)
-				Delegate.CreateDelegate(typeof(Func<IHandler[], IHandler>), instance, (MethodInfo)filterMethod);
+
+			MethodInfo methodInfo = filterMethod as MethodInfo;
```
jonorossi: This shouldn't use a safe-cast, otherwise the following line will just throw a null-ref exception.

alinapopa: Changed to direct cast.

jonorossi: Thanks

PR #266, opened 2017-05-26
Replace the methods that need Assembly.GetCallingAssembly()

alinapopa:
Require the calling assembly to be passed in, for methods that need `Assembly.GetCallingAssembly()`


src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
```diff
@@ -115,13 +115,21 @@ public static IWindsorInstaller InDirectory(AssemblyFilter filter, InstallerFact
...
 		/// </remarks>
+#if FEATURE_GETCALLINGASSEMBLY
 		[MethodImpl(MethodImplOptions.NoInlining)]
 		public static IWindsorInstaller InThisApplication()
 		{
 			var assembly = Assembly.GetCallingAssembly();
 			return ApplicationAssemblies(assembly, new InstallerFactory());
 		}
+#else
+		public static IWindsorInstaller InThisApplication(Assembly callingAssembly)
```
jonorossi: I asked below for the new argument to be named `assembly`, I think this one (`InThisApplication`) should be named `rootAssembly` since that is consistent with `ApplicationAssemblies` and calls out that it can't just be any assembly because it'll recursively follow referenced assemblies.

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
```diff
@@ -115,13 +115,21 @@ public static IWindsorInstaller InDirectory(AssemblyFilter filter, InstallerFact
 		///     <c>MyApp.SuperFeatures.dll</c>, <c>mscorlib.dll</c> and <c>ThirdPartyCompany.UberControls.dll</c> the <c>MyApp.exe</c> and <c>MyApp.SuperFeatures.dll</c> will be scanned for installers, and other
 		///     assemblies will be ignored.
 		/// </remarks>
+#if FEATURE_GETCALLINGASSEMBLY
```
jonorossi: After reading the comments through the corefx issues about not wanting to implement `GetCallingAssembly` because of problems with inlining that we haven't actually fully suppressed with `MethodImplOptions.NoInlining` (see quote below from MSDN about tail calls), I think the new methods that require an `Assembly` should be available on both .NET Framework and .NET Core builds, and the old ones using `GetCallingAssembly` should be wrapped in `FEATURE_GETCALLINGASSEMBLY` and kept for compatibility so if users run into problems with them they'll need to switch over to the new ones (if we do get problems in the future we'll want to mark them `[Obsolete]`). I am aware that the team gave in and has added `GetCallingAssembly` to .NET Standard 2.0, I still think this is the best way forward, and will significantly clean up the unit tests so they can all use the same API across both platforms.

> This effect also occurs when M1 executes as a tail call from M2, or when M2 executes as a tail call from M3. You can prevent the JIT compiler from inlining the method that calls GetCallingAssembly, by applying the MethodImplAttribute attribute with the MethodImplOptions.NoInlining flag, but there is no similar mechanism for preventing tail calls. (https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getcallingassembly(v=vs.110).aspx)

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
alinapopa: Changed to rootAssembly

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
```diff
@@ -139,6 +147,12 @@ public static IWindsorInstaller InThisApplication(InstallerFactory installerFact
 			var assembly = Assembly.GetCallingAssembly();
 			return ApplicationAssemblies(assembly, installerFactory);
 		}
+#else
+		public static IWindsorInstaller InThisApplication(InstallerFactory installerFactory, Assembly callingAssembly)
```
alinapopa: Switched the arguments

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
```diff
@@ -203,6 +218,17 @@ public static IWindsorInstaller This(InstallerFactory installerFactory)
 		{
 			return Instance(Assembly.GetCallingAssembly(), installerFactory);
 		}
+#else
+		public static IWindsorInstaller This(Assembly callingAssembly)
```
alinapopa: Removed the method

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
```diff
@@ -203,6 +218,17 @@ public static IWindsorInstaller This(InstallerFactory installerFactory)
...
+		{
+			return Instance(callingAssembly);
+		}
+
+		public static IWindsorInstaller This(InstallerFactory installerFactory, Assembly callingAssembly)
```
alinapopa: Removed the method

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
alinapopa: I've made changes so that methods that require an `Assembly` argument are available on both, and changed the argument name to `assembly`.
How about the tests? You've mentioned cleaning them up, meaning I assume that they all will use the new methods that pass the assembly in. But then there will be no coverage for the old methods.

src/Castle.Windsor/MicroKernel/Registration/Classes.cs
```diff
@@ -108,12 +108,20 @@ public static FromAssemblyDescriptor FromAssemblyInDirectory(AssemblyFilter filt
 		///     <c>MyApp.SuperFeatures.dll</c>, <c>mscorlib.dll</c> and <c>ThirdPartyCompany.UberControls.dll</c> the <c>MyApp.exe</c> and <c>MyApp.SuperFeatures.dll</c> will be scanned for components, and other
 		///     assemblies will be ignored.
 		/// </remarks>
+#if FEATURE_GETCALLINGASSEMBLY
 		[MethodImpl(MethodImplOptions.NoInlining)]
 		public static FromAssemblyDescriptor FromAssemblyInThisApplication()
 		{
 			var assemblies = new HashSet<Assembly>(ReflectionUtil.GetApplicationAssemblies(Assembly.GetCallingAssembly()));
 			return new FromAssemblyDescriptor(assemblies, Filter);
```
jonorossi: Could we change the implementation of this method to just call the new one below and pass `GetCallingAssembly()`.

src/Castle.Windsor/MicroKernel/Registration/Types.cs
```diff
@@ -108,12 +108,20 @@ public static FromAssemblyDescriptor FromAssemblyInDirectory(AssemblyFilter filt
 		///     <c>MyApp.SuperFeatures.dll</c>, <c>mscorlib.dll</c> and <c>ThirdPartyCompany.UberControls.dll</c> the <c>MyApp.exe</c> and <c>MyApp.SuperFeatures.dll</c> will be scanned for components, and other
 		///     assemblies will be ignored.
 		/// </remarks>
+#if FEATURE_GETCALLINGASSEMBLY
 		[MethodImpl(MethodImplOptions.NoInlining)]
 		public static FromAssemblyDescriptor FromAssemblyInThisApplication()
 		{
 			var assemblies = new HashSet<Assembly>(ReflectionUtil.GetApplicationAssemblies(Assembly.GetCallingAssembly()));
 			return new FromAssemblyDescriptor(assemblies, null);
```
jonorossi: Could we change the implementation of this method to just call the new one below and pass `GetCallingAssembly()`.

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
```diff
@@ -115,12 +115,19 @@ public static IWindsorInstaller InDirectory(AssemblyFilter filter, InstallerFact
 		///     <c>MyApp.SuperFeatures.dll</c>, <c>mscorlib.dll</c> and <c>ThirdPartyCompany.UberControls.dll</c> the <c>MyApp.exe</c> and <c>MyApp.SuperFeatures.dll</c> will be scanned for installers, and other
 		///     assemblies will be ignored.
 		/// </remarks>
+#if FEATURE_GETCALLINGASSEMBLY
 		[MethodImpl(MethodImplOptions.NoInlining)]
 		public static IWindsorInstaller InThisApplication()
 		{
 			var assembly = Assembly.GetCallingAssembly();
 			return ApplicationAssemblies(assembly, new InstallerFactory());
```
jonorossi: Could we change the implementation of this method to just call the new one below and pass `GetCallingAssembly()`.

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
jonorossi: Thanks, I've added a few more comments to those changes.

With regards to the unit tests good point, however I still think it would be best to standardise both platforms rather than heaps of extra conditional compilation.

Below are the 9 methods that use `GetCallingAssembly`, if they are just a single line that passes through the assembly to the implementation then unit testing them is much less important. `AllTypes` is already obsolete so I wouldn't worry about a unit test for that one. If we could duplicate a basic unit test for `Classes.FromThisAssembly()` and `FromAssembly.This()` both compiled only for .NET Framework I'd be happy for the rest to use the one that requires an assembly argument.

- `AllTypes.FromThisAssembly()` (calls `Classes.FromAssembly(assembly)`)
- `Classes.FromAssemblyInThisApplication()`
- `Classes.FromThisAssembly()`
- `Types.FromAssemblyInThisApplication()`
- `Types.FromThisAssembly()`
- `FromAssembly.InThisApplication()`
- `FromAssembly.InThisApplication(installerFactory)`
- `FromAssembly.This()`
- `FromAssembly.This(installerFactory)`

src/Castle.Windsor/MicroKernel/Registration/Classes.cs
alinapopa: Changed

src/Castle.Windsor/MicroKernel/Registration/Classes.cs
```diff
@@ -108,12 +108,20 @@ public static FromAssemblyDescriptor FromAssemblyInDirectory(AssemblyFilter filt
...
 		}
+#endif
+
+		public static FromAssemblyDescriptor FromAssemblyInThisApplication(Assembly assembly)
```
alinapopa: Changed name to rootAssembly

src/Castle.Windsor/MicroKernel/Registration/Classes.cs
```diff
@@ -130,11 +138,18 @@ public static FromAssemblyDescriptor FromAssemblyNamed(string assemblyName)
...
 		/// </returns>
+#if FEATURE_GETCALLINGASSEMBLY
 		[MethodImpl(MethodImplOptions.NoInlining)]
 		public static FromAssemblyDescriptor FromThisAssembly()
 		{
 			return FromAssembly(Assembly.GetCallingAssembly());
 		}
+#endif
+
+		public static FromAssemblyDescriptor FromThisAssembly(Assembly assembly)
```
alinapopa: Removed method

src/Castle.Windsor/MicroKernel/Registration/Types.cs
alinapopa: Changed

src/Castle.Windsor/MicroKernel/Registration/Types.cs
```diff
@@ -108,12 +108,20 @@ public static FromAssemblyDescriptor FromAssemblyInDirectory(AssemblyFilter filt
...
 		}
+#endif
+
+		public static FromAssemblyDescriptor FromAssemblyInThisApplication(Assembly assembly)
```
alinapopa: Changed name to rootAssembly

src/Castle.Windsor/MicroKernel/Registration/Types.cs
```diff
@@ -130,10 +138,17 @@ public static FromAssemblyDescriptor FromAssemblyNamed(string assemblyName)
...
 		/// </returns>
+#if FEATURE_GETCALLINGASSEMBLY
 		[MethodImpl(MethodImplOptions.NoInlining)]
 		public static FromAssemblyDescriptor FromThisAssembly()
 		{
 			return FromAssembly(Assembly.GetCallingAssembly());
 		}
+#endif
+
+		public static FromAssemblyDescriptor FromThisAssembly(Assembly assembly)
```
alinapopa: Removed the method

src/Castle.Windsor/MicroKernel/Registration/AllTypes.cs
```diff
@@ -102,11 +102,18 @@ public static FromAssemblyDescriptor FromAssemblyNamed(string assemblyName)
...
 		/// <returns>The corresponding <see cref = "FromDescriptor" /></returns>
+#if FEATURE_GETCALLINGASSEMBLY
 		[MethodImpl(MethodImplOptions.NoInlining)]
 		public static FromAssemblyDescriptor FromThisAssembly()
 		{
 			return Classes.FromAssembly(Assembly.GetCallingAssembly());
 		}
+#endif
+
+		public static FromAssemblyDescriptor FromThisAssembly(Assembly assembly)
```
alinapopa: Removed

src/Castle.Windsor/Windsor/Installer/FromAssembly.cs
alinapopa: Changed

jonorossi:
@alinapopa thanks for the changes, looks great.

PR #267, opened 2017-06-01
Add conditional compilation symbol FEATURE_APPDOMAIN

alinapopa:
Used the implementation from [Fortress ](https://github.com/cryosharp/fortress)for [Compatibility/ApplicationDomains ](https://github.com/cryosharp/fortress/blob/master/src/Fortress.Windsor/Compatibility/ApplicationDomains.cs) and for [GetLoadedAssemblies()](https://github.com/cryosharp/fortress/blob/master/src/Fortress.Windsor/Core/Internal/ReflectionUtil.cs#L114)


src/Castle.Windsor.Tests/Bugs/IoC-267.cs
```diff
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if FEATURE_APPDOMAIN
```
jonorossi: This probably should be `FEATURE_CODEDOM` which comes back in .NET Standard 2.0.

The line in the unit test using `AppDomain` could easily be changed to:
```
var coreAssembly = typeof(Castle.Core.ProxyServices).Assembly.Location;
```

src/Castle.Windsor/Compatibility/ApplicationDomains.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace Castle.Core.Internal
+{
+    internal static class ApplicationDomains
```
jonorossi: I'm thinking this would be better named `AppContext` and conditionally compiled only for .NET 4.5, so that when we rev to .NET 4.6 in the future this code doesn't have to be yanked out.

src/Castle.Windsor/Compatibility/ApplicationDomains.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public static string BaseDirectory
+        {
+            get
+            {
+#if FEATURE_APPDOMAIN
+                return  AppDomain.CurrentDomain.BaseDirectory;
+#else
+				return AppContext.BaseDirectory;
+#endif
```
jonorossi: Indentation.

src/Castle.Windsor/Core/Internal/ReflectionUtil.cs
```diff
@@ -150,14 +150,20 @@ public static Assembly GetAssemblyNamed(string assemblyName)
...
 			return AppDomain.CurrentDomain.GetAssemblies();
+#else
+			string basePath = AppContext.BaseDirectory;
+
+			var files = new DirectoryInfo(basePath).GetFiles("*.dll");
+			Assembly[] results = new Assembly[files.Length];
+			
+			for(int i = 0; i < files.Length; i++)
+			{
+				results[i] = System.Runtime.Loader.AssemblyLoadContext.Default.LoadFromAssemblyPath(files[i].FullName);
```
jonorossi: Good to see `AppDomain.CurrentDomain.GetAssemblies()` comes back in .NET Standard 2.0, we should just require 2.0 😄 .

However as this code stands eagerly loading all DLLs from the bin directory will cause problems, firstly performance by forcing them to load and secondly exceptions on native DLLs in there (e.g. libcoreclr.dll).

src/Castle.Windsor/Compatibility/ApplicationDomains.cs
alinapopa: Changed ApplicationDomains into AppContext. This file is duplicated into the Castle.Windsor.Tests project, because internals are not visible to tests. I will create a shared project in the solution, with AppContext.cs and NetCoreReflectionExtensions.cs.

src/Castle.Windsor/Core/Internal/ReflectionUtil.cs
alinapopa: It looks like there aren't any options to get these assemblies without loading them. The number of loaded assemblies can be reduced by filtering the names with `ShouldSkipAssembly()` before loading. Checking `AssemblyName.ContentType` can help filter out the native assemblies.
I have pushed a commit that uses `DependencyContext.RuntimeLibraries` to get the assemblies. It works with .NET Standard 1.6. To support .NET Standard 1.5, the way to get these assemblies would be still to look in the base directory for dll's, or if the application assembly is passed in, to get them with  `Assembly.GetReferencedAssemblies()`

src/Castle.Windsor/MicroKernel/SubSystems/Conversion/TypeNameConverter.cs
```diff
@@ -130,9 +134,10 @@ private Type GetType(string name)
 		private bool InitializeAppDomainAssemblies(bool forceLoad)
 		{
 			var anyAssemblyAdded = false;
+#if FEATURE_APPDOMAIN
 			if (forceLoad || assemblies.Count == 0)
 			{
-				var loadedAssemblies = ReflectionUtil.GetLoadedAssemblies();
```
fir3pho3nixx: I am not sure we should have done this. It was quite trivial to scan for assemblies eagerly by scanning the base directory.

Please see: https://github.com/cryosharp/fortress/blob/master/src/Fortress.Windsor/Core/Internal/ReflectionUtil.cs#L119

src/Castle.Windsor/MicroKernel/SubSystems/Conversion/TypeNameConverter.cs
```diff
@@ -144,6 +149,29 @@ private bool InitializeAppDomainAssemblies(bool forceLoad)
 					Scan(assembly);
 				}
 			}
+#else
+			if (assemblies.Count == 0)
+			{
+				var context = DependencyContext.Default;
```
fir3pho3nixx: Introducing a dependency to solve this problem is leaving me a little puzzled. What does it give us that we cannot do ourselves? (Excuse my ignorance).

alinapopa: @fir3pho3nixx I do not have preference for one option or another (see previous comments on this PR) but I think using the `DependencyContext` is more robust because it gives the real dependencies, versus whatever dll's may be in the bin folder. Also loading just the dll's from the base directory may miss some assemblies. I am seeing this test failing [`Throws_helpful_exception_when_assembly_found_but_not_type`](https://github.com/castleproject/Windsor/blob/netcore/src/Castle.Windsor.Tests/TypeNameConverterTestCase.cs#L215)
which looks in Castle.Core for a type, but Castle.Core.dll is not present in the bin folder of the tests, even though it is loaded.

fir3pho3nixx: I agree it will break for projects that consume statically compiled dll's via extern forwards but this is solved by adding error handling. I am not sure I know what you mean when you say Castle.Core is not in the bin folder for tests. It should be because of the NuGet reference. Why is it disappearing?

src/Castle.Windsor/Compatibility/ApplicationDomains.cs
jonorossi: I'd prefer them to be duplicated than a shared project, otherwise you could just include the 2 files from the Castle.Windsor directory.

src/Castle.Windsor/MicroKernel/SubSystems/Conversion/TypeNameConverter.cs
jonorossi: @fir3pho3nixx did you see my comments in https://github.com/castleproject/Windsor/pull/267#discussion_r119878886?

I've worked on some big .NET applications using Windsor before where the bin directory contained hundreds of DLLs, so eagerly loading all DLLs would massively slow down start up of those applications and increase memory usage. This code path could also be called quite a few times in `TypeNameConverter.GetType` and failed fusion binds throwing exceptions will make things messy.

Alternatively I'm not that impressed with using `Microsoft.Extensions.DependencyModel` either because it is designed for .NET CLI not .NET Core/Standard, it doesn't return the assemblies loaded into the CLR but the "runtime assemblies" specified in the `.deps` file for .NET CLI projects.

I guess our options are limited at this time without going with .NET Standard 2.0.

fir3pho3nixx: @jonorossi - I am also not completely sold on the dependency model stuff(+1). The case where there were hundreds of assemblies, is this quite common on the projects you work on? I have generally broken things up using services so never had this before. If this is common for you, then we should emulate this scenario somehow using a test. The worry here is we need to change this anyway, should we be concerned about how this executes under load?

jonorossi: > The case where there were hundreds of assemblies, is this quite common on the projects you work on?

Nope not common, however I've heard of many other projects with large bin directories, however most tend to be thick clients so not really a problem for .NET Core with no WinForms since people can't migrate those to .NET Core.

> The worry here is we need to change this anyway, should we be concerned about how this(dep model stuff) executes under load?

They are obviously both going to perform poorly compared to the .NET Framework implementation if they eager load all assemblies.  However at least the `DependencyModel` stuff knows which assemblies to load and that info is already loaded by the `dotnet` CLI runtime host.

The other option since .NET Standard 2.0 is bringing `AppDomain.CurrentDomain.GetAssemblies()` back is to just use reflection to call it since coreclr has always had the APIs by the looks of it (just not exposed by corefx), we've done this many times in the past with different .NET Framework versions, it sucks but it'll mean we can more easily transition to .NET Standard 2.0 and avoid the other problems.

I'm just throwing my thoughts out there, if you feel strongly towards one approach let's talk on it.

fir3pho3nixx: In that case I would opt for dropping the dependency and going for the low tech option I suggested earlier. The reason I say this is once the dependency is baked in, it's API usage would only increase over time proliferating throughout the code base. This should not be allowed to happen if there is an imminent solution. I can also only find this in it's current form from people that forked the cli ages ago on github. The current master branch of the cli does not appear to have this code anymore. There be dragons here. 

@alinapopa - Can you help out with this discussion? Where is this project currently maintained? Where has it moved to?

alinapopa: `DependencyContext` source is at https://github.com/dotnet/core-setup/blob/master/src/managed/Microsoft.Extensions.DependencyModel/DependencyContext.cs

eerhardt: @alinapopa asked me to help clarify some questions about DependencyModel.  I am more or less the maintainer of it - see [here](https://github.com/dotnet/core-setup/tree/master/src/managed/Microsoft.Extensions.DependencyModel) for the latest code and issues, etc.

I'm not really going to recommend whether you use it one way or the other, but I just wanted to give information and clear up some questions.


To start, my 5 second intro to the DependencyModel library is that it is the managed object model for the `.deps.json` file that is created as part of every .NET Core application.  The `dotnet.exe` native host uses the .deps.json file to know which assemblies your app requires and which it should load.  Managed code can also retrieve this information to know about the dependencies of the application.

Currently, I'm aware of 2 "high profile" places that uses the DependencyModel:
1. [ASP.NET MVC](https://github.com/aspnet/Mvc/blob/c8a1bb69142792d1046044bde7082b4e3ac8ac27/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultAssemblyPartDiscoveryProvider.cs#L38) apps use it to discover "Controller" types in the application.
1. The [VS test runner for xunit](https://github.com/xunit/xunit/blob/5133b76c1add95351bef471a6900de6155294370/src/xunit.runner.visualstudio/VsTestRunner.cs#L618) uses it to find if the test references any "runner reporter" assemblies.  

Given it is used by things like the ASP.NET MVC runtime, it would be a HUGE breaking change if we ever stopped producing/maintaining this library.  So I wouldn't be worried about the support for it going away anytime soon.  Also, from the way it sounds, you would only be using this library as a stop-gap to use in .NET Core 1.x applications.  Once you are on `netstandard2.0`, you can use `AppDomain.CurrentDomain.GetAssemblies()` again.

> Alternatively I'm not that impressed with using Microsoft.Extensions.DependencyModel either because it is designed for .NET CLI not .NET Core/Standard

> The current master branch of the cli does not appear to have this code anymore.

The DependencyModel library moved "lower" in the stack from the dotnet/cli repo to dotnet/core-setup - the same repo where the `dotnet.exe` host is maintained.  It was recognized that this library is useful at runtime and not just as part of the "tooling".  It is maintained by the same team that maintains the dotnet/corefx and dotnet/coreclr repos.

> The other option since .NET Standard 2.0 is bringing AppDomain.CurrentDomain.GetAssemblies() back is to just use reflection to call it since coreclr has always had the APIs by the looks of it (just not exposed by corefx),

Using reflection doesn't work on `netcoreapp1.x` because the `GetAssemblies` method was removed in a post-build step since it isn't called internally.  This is why we don't recommend approaches like this.  You can see for yourself if you JustDecompile/ildasm the `System.Private.CoreLib` assembly that ships with `netcoreapp1.x`.  ex. `C:\Program Files\dotnet\shared\Microsoft.NETCore.App\1.1.2\System.Private.CoreLib.dll`.

### Other Thoughts
Using the DependencyModel has an advantage that it gives you a graph of the dependencies.  This is useful if you are searching for types that implement a certain interface, or inherit from a certain class.  You can immediately dismiss any assemblies that don't depend on your core library.  You can see this approach being taken in the ASP.NET MVC example above.  They don't need to inspect assemblies like `Newtonsoft.Json` to see if it contains any MVC Controller classes.  It can't because it doesn't reference the MVC library.

If necessary, you can only depend on the DependencyModel from the `netstandard1.x` build of your project, which means its API usage would be harder to proliferate over time because it would always needed to be `#if` checked.  And when you moved to support `netstandard2.0`, the dependency could be removed, so it would only be supported on the `netstandard1.x` version of your project.

src/Castle.Windsor/Castle.Windsor-VS2017.csproj
```diff
@@ -35,4 +35,10 @@
   <PropertyGroup Condition="'$(TargetFramework)'=='net45'">
     <DefineConstants>$(DefineConstants);FEATURE_PERFCOUNTERS;FEATURE_GAC</DefineConstants>
   </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard1.5'">
```
eerhardt: Where does this target framework come from?  Above, this project only declares:

`<TargetFramework>net45</TargetFramework>`

fir3pho3nixx: @eerhardt - Thanks for the explanation. 

> Where does this target framework come from? Above, this project only declares:
> <TargetFramework>net45</TargetFramework>

The intention is to multi-target similar to https://github.com/castleproject/Core/blob/refactor-build/src/Castle.Core/Castle.Core.csproj#L6 eventually. 

I am sure you can appreciate things are moving around a fair bit here, time for us is also at a premium and "chasing moved cheese" becomes rather dull after a while(sorry if this offends). My challenge was merely to understand why a dependency that potentially could be superseded by roughly 7-8 lines of code was nowhere to be found on the cli repo when I knew it was there before. This got me round to thinking is this really another "goose chase" we want at this stage. 

Your explanation has definitely helped. I will bow out of this PR and leave it to @jonorossi and @alinapopa to decide whether to take it forward.

Again thanks for taking the time to explain although I still believe we don't need it for most use cases.

alinapopa: @eerhardt The targets are not set because this is work in progress and the project does not build on core yet. The targets should be `<TargetFrameworks>net45;netstandard1.6</TargetFrameworks>`
I have a version that builds with the right targets at https://github.com/alinapopa/Windsor/tree/netcoreAll
Also the `netstandard1.5` condition is there because I initially planned the target to be netstandard1.5; if the `DependencyModel` will be used, the target will be changed to 1.6.

jonorossi:
Many thanks @eerhardt, that gives me a much better idea of how this little library is maintained and used.

At this stage I'm just going to make the decision to go with the DependencyModel library, it allows us to skip assemblies by their assembly name without even loading them, possibly inspect dependencies if we wanted to improve the current dependency lookup, means we won't get fusion bind failures from .dll files we shouldn't even try loading (e.g. native runtime DLLs), and is more correct with ASP.NET MVC and the VS test runner using it so it won't just stop working. @fir3pho3nixx, I understand your reluctance to go this way but I think it'll be the best short term solution for this missing API.

With that I'm going to merge this pull request as the work is already done. Thanks everyone for contributing to this pull request, let's try to move on to the remaining work.

PR #268, opened 2017-06-06
Replace AssemblyName.GetAssemblyName() and Assembly.Load(string)

alinapopa:
Added FEATURE_ASSEMBLIES for conditional compilation of AssemblyName.GetAssemblyName() and Assembly.Load(string), and used AssemblyLoadContext methods on Core. Both these API's are coming back in .NET Standard 2.0


jonorossi:
Looks great, thanks.

Issue #241, opened 2017-05-17
Set up build for .NET Core


alinapopa:
Using AppVeyor and Travis CI, similar to Castle.Core - see Castle.Core issue [241](https://github.com/castleproject/Core/issues/241)

fir3pho3nixx:
@alinapopa && @jonorossi - How do we want to do this? 

All the `refactor-build` code is merged into `netcore` branch.

Should I just enable tests for netcoreapp in AppVeyor and create a failed build using a PR(netcore-tests -> netcore)? 

Also do we want an issue for getting mono rocking? We had this before, I would like to bring it back. I just need guidance from you as to where and how we place this.

alinapopa:
> Should I just enable tests for netcoreapp in AppVeyor 

Yes I think this is needed.

jonorossi:
> All the refactor-build code is merged into netcore branch.

Is there much point in keeping the refactor-build branch, we are so close to getting a .NET Standard build going we aren't likely to release anything without that now?

> Should I just enable tests for netcoreapp in AppVeyor and create a failed build using a PR(netcore-tests -> netcore)?

I agree with @alinapopa, yes.
 
> Also do we want an issue for getting mono rocking? We had this before, I would like to bring it back on Travis. I just need guidance from you as to where and how we place this.

We didn't have this for Windsor before only Castle Core, however since you've been able to get it working pretty cleanly for Castle Core I don't see why not adding it and working through any issues while we sort out the remaining bits for .NET Standard.

fir3pho3nixx:
Cool, will get this tests failing on a new branch called `netcore-tests` and raise a PR to `netcore` so we can start looking at the failures and raising more issues. I will also raise a separate issue for mono but leave it out of the `ready` column until we have made more progress with the tests. 

👍

fir3pho3nixx:
We now have a good build in `netcore` and 3 failing tests along with 1 ignored test in `netcore-tests`. Looking good people! #277

jonorossi:
@fir3pho3nixx what's the plan for restoring `WillRegisterServiceWithServiceCatalog` in `src/Castle.Facilities.WcfIntegration.Tests/Service/ServiceHostFixture.cs` which you marked ignored?

fir3pho3nixx:
Sorry was looking into the dotnet test issue on Linux. The vstest team has scheduled a fix for 15.3 milestone. 

I think we should fix it. Will take another look when I am back from hols next week.

[...]

Issue #237, opened 2017-05-17
Add .editorconfig


alinapopa:

jonorossi:
I've copied the editorconfig from Core in 39216ec9dbaf7069eeb8a3dda0fc8b4c51d960b4.

PR #270, opened 2017-06-07
Netcore refactor build

fir3pho3nixx:
I have updated the netcore branch with the refactor-build changes. 

I did run into some issues though. We have 2 versions of the AppContext in the compatibility layer, I deleted the one in the test project because I thought it would be good to leverage `InternalsVisibleTo` just like Core does to get round the `internal` class modifier. Sorry if I missed some conversations on this(I blame British Telecoms). 

We also have a few tests failing failing, not too serious mind you but I thought I would start a conversation and get the PR going. 

My goal here is to get a failed build going for netcore so we can start migrating tests. We also need to consider some of the SILVERLIGHT build symbols still knocking about.


fir3pho3nixx:
#241

alinapopa:
For the AppContext there was a discussion https://github.com/castleproject/Windsor/pull/267#discussion_r120069735
I have included the file from Castle.Windsor directory in this commit: https://github.com/castleproject/Windsor/pull/272/commits/1f8f2fe3520f709a19b298639620df361139c7ca

fir3pho3nixx:
@alinapopa - Interesting. I still had to add these namespaces to make it compile after deleting it. 

https://github.com/castleproject/Windsor/pull/270/commits/6c622e46481b25de87797f421a12ca17d8539042#diff-8cd94cbc9518ce2b6b4a83f3b20a4c2fR21

https://github.com/castleproject/Windsor/pull/270/commits/6c622e46481b25de87797f421a12ca17d8539042#diff-91149d79f52b6063cd5d4a9b37aa963fR21

https://github.com/castleproject/Windsor/pull/270/commits/6c622e46481b25de87797f421a12ca17d8539042#diff-13278852499736e2295dc81caae6f232R22

https://github.com/castleproject/Windsor/pull/270/commits/6c622e46481b25de87797f421a12ca17d8539042#diff-a5c21b4a91aa747c7cfdf8471836def6R22

https://github.com/castleproject/Windsor/pull/270/commits/6c622e46481b25de87797f421a12ca17d8539042#diff-79045564f824b06efee7be484c993bdcR24

https://github.com/castleproject/Windsor/pull/270/commits/6c622e46481b25de87797f421a12ca17d8539042#diff-426a283953c953ce976a76f3e6716567R23

fir3pho3nixx:
@alinapopa - We need to sync up and get on the same branch.

fir3pho3nixx:
@alinapopa - We can then add [this](https://github.com/castleproject/Core/blob/refactor-build/buildscripts/build.cmd#L56) to the build script and the framework moniker for netstandard to the [test project](https://github.com/castleproject/Windsor/blob/netcore/src/Castle.Windsor.Tests/Castle.Windsor.Tests-VS2017.csproj#L4). This would yield failed tests in pursuit of #271.

fir3pho3nixx:
@alinapopa 

> For the AppContext there was a discussion #267 (comment)

I think what @jonorossi was saying here was he was not happy using a shared project. What I have also seen used before was the use of the [InternalsVisibleTo](https://github.com/castleproject/Core/blob/refactor-build/src/Castle.Core/Properties/InternalsVisibleToTests.cs#L17) attribute to expose internals to tests which to me seemed like a tidier way of achieving this. 

Will ping back tomorrow after your feedback.

src/Castle.Windsor/Castle.Windsor-VS2017.csproj
```diff
@@ -1,41 +0,0 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
-
-  <PropertyGroup>
-    <TargetFrameworks>net45;netstandard1.6</TargetFrameworks>
-  </PropertyGroup>
-  
-  <Import Project="..\..\common.props"></Import>
```
alinapopa: Please keep this order in all csproj files, first target frameworks then import of common.props.

src/Castle.Windsor/Castle.Windsor.csproj
```diff
@@ -1,754 +1,41 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\buildscripts\common.props"></Import>
```
alinapopa: The import line needs to be after the target frameworks - see original csproj file. The reason is that precompile symbols (defined in common.props) depend on the target framework. You can verify that by opening the solution in Visual Studio and look for any symbol that should be defined on net45, such as `#if FEATURE_REMOTING` and you'll see that the active/inactive code looks wrong. This is also why the tests fail.

fir3pho3nixx: > This is also why the tests fail.

Never noticed this ordering causing tests to fail in Castle Core or Fortress. 

The code highlighting thing is interesting. I just thought that was VS getting confused over multi-target projects. If this is the fix for it, will happily make the change. Thanks.

Castle.Windsor.sln.DotSettings
```diff
@@ -16,6 +16,7 @@
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/XmlDocFormatter/WrapBeforeAttr/@EntryValue">False</s:Boolean>
```
fir3pho3nixx: Isn't this how you share re-sharper settings?

src/Castle.Windsor/Compatibility/AppContext.cs
```diff
@@ -17,7 +17,10 @@
 #if DOTNET45
 namespace Castle.Core.Internal
 {
-    internal static class AppContext
+	// This should be changed to expose externals similar to how core does it. Rather then copy this down to tests we should 
+	// be using the InternalsVisibleTo attribute. Will raise this on the PR in GitHub. Making it public for now to get
+	// the project to compile. If you are reviewing this, please flag it.
+    public static class AppContext
```
jonorossi: I think I just reviewed @alinapopa get rid of the duplicate class in the other open PR.

jonorossi:
> I think what @jonorossi was saying here was he was not happy using a shared project. What I have also seen used before was the use of the InternalsVisibleTo attribute to expose internals to tests which to me seemed like a tidier way of achieving this.

Yep, I thought @alinapopa was talking about those "shared project" types that were the alternative to PCLs that don't compile but just hold code compiled into something else.

`InternalsVisibleTo` might not be the best option, it might cause you grief in the future if you accidentally don't expose something needed for normal use of Windsor but won't be picked up by unit tests because they could then use internal code.

Since we are literally talking about one single line method that is a polyfill I didn't see a problem being duplicated between projects.

src/Castle.Windsor/Castle.Windsor.csproj
alinapopa: The tests failed because FEATURE_REMOTING was expected to be defined in the tests project and it was not, so the test execution took a different path. Changing the ordering is a workaround to fix it.
For some reason this happened in the test project, but not on main Windsor project.

alinapopa:
@fir3pho3nixx I won't be around until much later today, feel free to merge after changing the ordering in csproj file so that tests pass.

fir3pho3nixx:
@alinapopa - thanks, will get these changes in by tomorrow evening. I am off to go vote this evening.

fir3pho3nixx:
> InternalsVisibleTo might not be the best option, it might cause you grief in the future if you accidentally don't expose something needed for normal use of Windsor but won't be picked up by unit tests because they could then use internal code.

Sorry @jonorossi but doing the right thing is difficult sometimes. I am siding with @kkozmic on this one. Duplicating type definitions across modules creates other cody type issues, eg the namespaces have to be the same for a start(with shared projects you might have this), behaviours can evolve independently over time if they are copied classes. If they are linked via msbuild, you are one change away from this going horribly wrong. I feel very strongly that this is a tidy way of hiding internals from your public api but creating a backdoor for testing without esoteric ephemeral linking or duplication. It also creates all sorts of mental issues when you start reflecting into assemblies to solve production problems. I am pretty sure this is OK.

fir3pho3nixx:
@alinapopa && @jonorossi - I am going to go ahead and merge this. In the absence of a positive signal about the InternalsVisibleTo attribute I have left this out for now. 

Thanks for reviewing.

fir3pho3nixx:
As an aside, I really enjoyed reading this: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/1709

jonorossi:
> I am going to go ahead and merge this.

👍 

> In the absence of a positive signal about the InternalsVisibleTo attribute I have left this out for now.

Oops, I didn't realise you were expecting a response. I said "InternalsVisibleTo might not be the best option" as that thought came to mind, however you clearly want to so let's go with it. I've used InternalsVisibleTo in the past for the exact same reason and it is a good reason, I just wanted a thought I had captured. Maybe doing this will help us get rid of those "Internal" namespaces.

> As an aside, I really enjoyed reading this: DefinitelyTyped/DefinitelyTyped#1709

Castle is tabs, and has been since 2003 😄, unfortunately there is a bit of mixed code that has snuck through. We've got #237 logged to add an editorconfig file like we've had in Core for ages.

fir3pho3nixx:
> Oops, I didn't realise you were expecting a response. I said "InternalsVisibleTo might not be the best option" as that thought came to mind, however you clearly want to so let's go with it. I've used InternalsVisibleTo in the past for the exact same reason and it is a good reason, I just wanted a thought I had captured. Maybe doing this will help us get rid of those "Internal" namespaces.

No problems at all. I will do it as a separate PR.

PR #279, opened 2017-06-11
Appveyor fixes so tag deploys work

fir3pho3nixx:


buildscripts/common.props
```diff
@@ -1,49 +1,48 @@
 <Project>
-  
-  <PropertyGroup>
-    <BuildVersion>3.4.0</BuildVersion>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
+
+	<PropertyGroup>
+		<BuildVersion>3.4.0</BuildVersion>
```
jonorossi: Can we remove this line.

Issue #273, opened 2017-06-08
Document all conditional compilation symbols


jonorossi:
We really need to document all the symbols like I did with Castle Core while they are fresh in our minds with a description of what is missing or different so we've got something to point people to as an overview of what is different and also so we can easily work out what we can remove as new APIs light up in .NET Standard in the future.

jonorossi:
Thanks @alinapopa.

PR #284, opened 2017-06-15
Document conditional compilation symbols

alinapopa:
#273


jonorossi:
I've made some changes to descriptions and fixed the h3 under an existing h2. Merging.

PR #288, opened 2017-06-28
Merge netcore branch to master

fir3pho3nixx:
Merging to master so we can deploy it.

Any final checks we would like to make or add here?


jonorossi:
I've just diffed the public API and output nuspecs against 3.4.0 and everything looks good.

alinapopa:
I would remove the symbol `DOTNET45`. There is a place that needs it but can be replaced with `NET45`

jonorossi:
> I would remove the symbol DOTNET45. There is a place that needs it but can be replaced with NET45

We can fix that after we merge, there are still a few things left to sort out before releasing.

jonorossi:
Great works guys, not long yet.

Issue #297, opened 2017-07-02
Remove single use of NETCOREAPP1_0 symbol


jonorossi:
`NETCOREAPP1_0` got used once in the codebase, let's get rid of that and the comment about Silverlight.

https://github.com/castleproject/Windsor/blob/27c2edd9d19f0f99263be2daa044544576c5bdfa/src/Castle.Windsor.Tests/Configuration2/EvalSupportTestCase.cs#L15

jonorossi:
Thanks for fixing this in #298. I logged it so I wouldn't forget, I didn't even have to fix it 👍

PR #298, opened 2017-07-02
Removed netcore compilation symbol as per #297

fir3pho3nixx:


fir3pho3nixx:
Interesting that a test to do with Wcf would fail from this change. 

https://ci.appveyor.com/project/castleproject/windsor/build/0.0.102/tests

but it passes when I merge it to master on my fork.

https://ci.appveyor.com/project/fir3pho3nixx/windsor/build/1.0.30/tests

fir3pho3nixx:
@jonorossi - Should we raise an issue and perhaps trigger the build manually again?

jonorossi:
> System.ServiceModel.AddressAlreadyInUseException : The TransportManager failed to listen on the supplied URI using the NetTcpPortSharing service: the URI is already registered with the service.

I haven't seen that error before, however if we get it again on our build server we should investigate.

I've re-run the build.

Issue #133, opened 2016-05-19
Documentation for PerWebRequest


altearius:
The documentation for the `PerWebRequest` life style notes that it is necessary to register the `PerRequestLifestyle` HTTP module to safely use this life style: https://github.com/castleproject/Windsor/blob/master/docs/lifestyles.md

It looks like this has not been necessary since version 3.0.

Perhaps the documentation should be updated to note this.

jonorossi:
@altearius yes, as long as `Microsoft.Web.Infrastructure` is available Windsor's [`PreApplicationStartMethod`](https://github.com/castleproject/Windsor/blob/c7a675659c9a426670c96d9f006f53f056705f67/src/Castle.Windsor/AutoregisterPerWebRequestLifestyleModule.cs) should get called therefore making manually registering the HTTP module unnecessary. Feel free to send through a PR to update the docs, thanks.

jonorossi:
This has been fixed in #320.

Issue #45, opened 2014-02-13
use HttpApplication.RegisterModule in .NET 4.5 for PerWebRequest lifestyle


kkozmic:
.NET 4.5 added a built in method so we no longer need to use `Microsoft.Web.Infrastructure.dll` for that in our `PerWebRequestLifestyleModuleRegistration`

http://msdn.microsoft.com/en-us/library/system.web.httpapplication.registermodule(v=vs.110).aspx

PR #320, opened 2017-08-25
Corrected documentation for PerWebRequest

ankitvijay:
Corrected documentation for PerWebRequest lifestyle. Refer https://github.com/castleproject/Windsor/issues/133

### DIFF
--- a/docs/lifestyles.md
+++ b/docs/lifestyles.md
@@ -47,8 +47,7 @@ This is how you register a component as per web request:
 ```csharp
 Container.Register(Component.For<MyPerWebRequestComponent>().LifestylePerWebRequest());
 ```
-
-:warning: **Registering `PerWebRequestLifestyleModule`:** In order to function properly per web request requires a `IHttpModule` - `Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule` to be registered in your web.config:
+:warning: **Registering `PerWebRequestLifestyleModule`:** (Only required if you are **NOT** running version 3.3 or above) In order to function properly per web request requires a `IHttpModule` - `Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule` to be registered in your web.config:
 
 ```xml
 <httpModules>
--- a/docs/lifestyles.md
+++ b/docs/lifestyles.md
@@ -47,25 +47,6 @@ This is how you register a component as per web request:
 ```csharp
 Container.Register(Component.For<MyPerWebRequestComponent>().LifestylePerWebRequest());
 ```
-:warning: **Registering `PerWebRequestLifestyleModule`:** (Only required if you are **NOT** running version 3.3 or above) In order to function properly per web request requires a `IHttpModule` - `Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule` to be registered in your web.config:
-
-```xml
-<httpModules>
-   <add name="PerRequestLifestyle" type="Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule, Castle.Windsor"/>
-</httpModules>
-```
-
-if you're running on IIS7 you may need to register it in `system.webServer/modules` section instead.
-
-```xml
-<configuration>
-   <system.webServer>
-      <modules>
-         <add name="PerRequestLifestyle" type="Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule, Castle.Windsor" />
-      </modules>
-   </system.webServer>
-</configuration>
-```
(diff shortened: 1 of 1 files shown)


ankitvijay:
Would like to understand why the checks have failed. It is a simple text change.

jonorossi:
Thanks for your pull request, looking at the build log it appears that either something has either changed with the AppVeyor OS image or some package wasn't deterministic and has broken since .NET Core 2.0 has been released. I've rebuild master and get the same error.

> NUnit.Engine.NUnitEngineException: The path specified in --result src\Castle.Windsor.Tests\bin\Release\net45\TestResult_Windsor.xml could not be written to ---> System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\projects\windsor\src\Castle.Windsor.Tests\bin\Release\net45\TestResult_Windsor.xml'.

docs/lifestyles.md
```diff
@@ -47,8 +47,7 @@ This is how you register a component as per web request:
 ```csharp
 Container.Register(Component.For<MyPerWebRequestComponent>().LifestylePerWebRequest());
 ```
-
-:warning: **Registering `PerWebRequestLifestyleModule`:** In order to function properly per web request requires a `IHttpModule` - `Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule` to be registered in your web.config:
+:warning: **Registering `PerWebRequestLifestyleModule`:** (Only required if you are **NOT** running version 3.3 or above) In order to function properly per web request requires a `IHttpModule` - `Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule` to be registered in your web.config:
```
jonorossi: I think it would be better to refer to the dependance on `Microsoft.Web.Infrastructure` for this to happen automatically as per #133.

This functionality also arrived in 3.0.0.

ankitvijay: I'm not very sure if it arrived in 3.0.0. Going through release notes, it looked like to come in [version 3.3.0 ](https://github.com/castleproject/Windsor/releases/tag/v3.3). Please refer [issue#45](https://github.com/castleproject/Windsor/issues/45) which was added as improvement in version 3.3.0. 

Additionally, should I also mention that it will work without any configuration changes from .NET Framework 4.5 and above?

jonorossi: 7a08d1a39b2828aa90f97956fbb8f7504ce272dc added the `PreApplicationStartMethod` attribute using the new `PerWebRequestLifestyleModuleRegistration` class for 3.0.0. 86d397c59fe467120afb45b8526b2a44f5c6f4d1 (from pull request #48) changed the reflection code we had that was using internals of `Microsoft.Web.Infrastructure` to use `HttpApplication.RegisterModule` under .NET 4.5, and the previous reflection code under .NET 3.5/4.0. Neither of us were totally right, from Windsor 3.0.0 it would be automatic but how it was automatic changed in 3.3.0 for .NET 4.5.

Windsor 4.x dropped support for .NET 3.5/4.0, however I guess people are still using the 3.x versions.

It looks like the exception Windsor throws if you haven't registered the module is a little outdated, i.e. referring to MVC3 and WebMatrix. I'm not sure how you obtain the M.W.I DLL now, however we don't even need that doco anymore since we only use `HttpApplication.RegisterModule` now since we only support .NET 4.5.

https://github.com/castleproject/Windsor/blob/e92d65fd8bce80c7617f3cd327552177cca9c50b/src/Castle.Windsor/MicroKernel/Lifestyle/PerWebRequestLifestyleModule.cs#L83-L107

Since you are actually using this functionality and obviously saw our docs are wrong I'm going to leave it up to you to decide what you think is worthwhile writing, however I personally see little value in mentioning how to use Windsor on .NET 3.5/4.0 in our docs because no one should be writing new code for those runtimes. It would be great if you could update the exception message in `PerWebRequestLifestyleModule` too.

Thanks

ankitvijay: I did not get any error message, the reason I got confused was because documentation said following:

> In order to function properly per web request requires a IHttpModule - Castle.MicroKernel.Lifestyle.PerWebRequestLifestyleModule to be registered in your web.config

It did not talk about **what and how error** would come, hence I assumed that it may lead to some unexpected error if I do not add this line.

I understand and absolutely agree with you we that docs should be updated for the latest version, hence we should remove this warning altogether. I will update this PR, removing the warning.

jonorossi:
Thanks @ankitvijay, sorry for the delay. Merged.

ankitvijay:
No problems @jonorossi , thanks for the merge

Issue #327, opened 2017-09-11
Master build is broken: could not load logging services because assembly versions don't match


jonorossi:
> Castle.MicroKernel.SubSystems.Conversion.ConverterException : Could not convert string 'Castle.Services.Logging.Log4netIntegration.ExtendedLog4netFactory,Castle.Services.Logging.Log4netIntegration,Version=4.0.0.0, Culture=neutral,PublicKeyToken=407dd0808d44fbdc' to a type.
>   ----> System.IO.FileLoadException : Could not load file or assembly 'Castle.Services.Logging.Log4netIntegration, Version=4.0.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

I've updating Windsor to use Castle Core 4.1.0, however because of a defect in the 4.x Castle Core build scripts (see https://github.com/castleproject/Core/issues/288) the assembly version contains too much.

We need to get https://github.com/castleproject/Core/issues/288 fixed and released, and bump the minimum version again otherwise consumers of Windsor will run into these errors on .NET Framework, we don't want users to have to deal with binding redirects.

fir3pho3nixx:
I am working on a fix for this. Hard coded assembly strings in logging facility ... yowser! 

288 is done in Core. I am also implementing it down here in the same PR. Should have fix for this soon.

fir3pho3nixx:
The logging facility is definitely something that is useful but the problem I find is that many different assembly versions converge on a single piece code which does "magic string" based invocations of inheritors of `Castle.Core.Logging.AbstractExtendedLoggerFactory`. 

The culprits are: 

 - Castle.Services.Logging.Log4netIntegration.ExtendedLog4netFactory
 - Castle.Services.Logging.NLogIntegration.ExtendedNLogFactory
 - Castle.Services.Logging.Log4netIntegration.Log4netFactory
 - Castle.Services.Logging.NLogIntegration.NLogFactory 

The problem we got is that subsequent releases of Castle.Core will almost always cause the Castle.Facilities.Logging to break. Looking at the history for [LoggingFacility](https://github.com/castleproject/Windsor/blob/bbc1183a8806f4cfdf07dc462e1895313fb0684a/src/Castle.Facilities.Logging/LoggingFacility.cs#L31) I can see that changing these "magic strings" is a thing for Castle Windsor releases:

[...]

PR #330, opened 2017-09-12
Forever fix core releases (v5)

fir3pho3nixx:
@jonorossi 

If we merge this and release a new major version of Castle.Windsor before deploying Castle.Core, then we wont ever have to worry about release drama with the "magic strings" in [LoggingFacility](https://github.com/castleproject/Windsor/blob/master/src/Castle.Facilities.Logging/LoggingFacility.cs#L33)

The magic strings have been traded out for some new facilities with strong typing and proper NuGet references:

[Castle.Facilities.Logging.NLog](https://github.com/fir3pho3nixx/Windsor/blob/facilities-logging-fix-deps/src/Castle.Facilities.Logging.NLog/LoggingFacility.cs#L268)

[Castle.Facilities.Logging.Log4net](https://github.com/fir3pho3nixx/Windsor/blob/facilities-logging-fix-deps/src/Castle.Facilities.Logging.log4net/LoggingFacility.cs#L268)

This has also highlighted the fact that these are not yet netcore compliant because their respective service counterparts in Castle.Core have not been upgraded yet. This was slightly hidden and not very obvious.

Personally, I am all for smooth releases what do you think?


fir3pho3nixx:
#327

jonorossi:
> then we wont ever have to worry about release drama with the "magic strings"

Closing this pull request because of my comment https://github.com/castleproject/Windsor/issues/327#issuecomment-329258548 and #336.

> This has also highlighted the fact that these are not yet netcore compliant because their respective service counterparts in Castle.Core have not been upgraded yet. This was slightly hidden and not very obvious.

* log4net (https://github.com/castleproject/Core/issues/201) - remember we ran into those `ILoggerRepository` problems since the API has less methods on .NET Core
* NLog (https://github.com/castleproject/Core/issues/200) - NLog still doesn't have a non prerelease for .NET Standard

fir3pho3nixx:
Reviewing your comments and then will review the PR. No problems! Like where this is going :)

Issue #324, opened 2017-09-11
Child container doesn't dispose all its resolved components on disposal


reijerh:
More specifically: child containers don't seem to dispose components that were resolved through _them_, but were registered in the _parent_ container.

Original issue: https://github.com/castleproject/Windsor/issues/27
It seems the core issue this ticket was about wasn't really fixed, at least the bug is still there in 4.0.

Here's a unit test (adopted from @mickdelaney's fork https://github.com/mickdelaney/Windsor/commit/7e8ab353e701918e6649cfb4bd22e6581ba4755a):

```
[Test]
public void DisposeChildContainerComponentsResolvedAgainstParentContainer()
{
	var parentContainer = new WindsorContainer();

	var childContainer = new WindsorContainer();
	parentContainer.AddChildContainer(childContainer);
	parentContainer.Register(Component.For<DisposableFoo>().LifestyleTransient());

	var first = childContainer.Resolve<DisposableFoo>();
	Assert.IsNotNull(first);
	childContainer.Release(first);
	Assert.AreEqual(1, DisposableFoo.DisposedCount);

	var second = childContainer.Resolve<DisposableFoo>();
	Assert.IsNotNull(second);
	childContainer.Dispose();
	Assert.AreEqual(2, DisposableFoo.DisposedCount);
}
```
Output:
>   Expected: 2
>  But was:  1

(NB: the original unit test registered the component as Singleton, meaning the test would already fail at the first case)

It seems that if we register the `DisposableFoo` at the `childContainer` instead, the object is disposed correctly.

You can see in the debugger that the object is tracked by the `parentContainer` rather than by the `childContainer`.

fir3pho3nixx:
@reijerh - This one was a tough nut to crack but I think I might have fixed it. 

Your test is [here](https://github.com/castleproject/Windsor/pull/335/commits/0666d64486933043a9e653f95e6038f3b5647b65#diff-c7e03dbd55f13e8f320644bfaf5e6490R126) and it now passes.

reijerh:
Thanks for the very quick response! I couldn't figure it out myself as an outsider due to the complex nature of the project. Will try out the fix when a new version is released.

fir3pho3nixx:
@reijerh - No problem. 

Once @jonorossi reviews the PR and we clean it up a little we will get this deployed for you.

jonorossi:
The problem here is that that component is only indirectly resolved via the child container, the component is registered, resolved, tracked and should be released in the parent container.

[...]

PR #335, opened 2017-09-13
Child container disposals

fir3pho3nixx:
Potential fix for #324

### DIFF
--- a/src/Castle.Windsor/MicroKernel/DefaultKernel.cs
+++ b/src/Castle.Windsor/MicroKernel/DefaultKernel.cs
@@ -392,6 +392,7 @@ namespace Castle.MicroKernel
 			}
 
 			var handler = NamingSubSystem.GetHandler(service);
+
 			if (handler == null && Parent != null)
 			{
 				handler = WrapParentHandler(Parent.GetHandler(service));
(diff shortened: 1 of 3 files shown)


fir3pho3nixx:
Rebased with master

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -725,7 +726,14 @@ protected virtual IHandler WrapParentHandler(IHandler parentHandler)
 				return null;
 			}
 
-			var handler = new ParentHandlerWrapper(parentHandler, Parent.Resolver, Parent.ReleasePolicy);
+			// If we are disposing child containers with parent registered transients, 
+			IReleasePolicy releasePolicy;
+			if (parentHandler.ComponentModel.LifestyleType == LifestyleType.Singleton || parentHandler.ComponentModel.LifestyleType == LifestyleType.Undefined)
+				releasePolicy = Parent.ReleasePolicy;
+			else
+				releasePolicy = this.ReleasePolicy;
```
jonorossi: This will break many scenarios, if the component was registered and resolved by the parent container it might not be a transient, what about pooled, scoped or any custom lifestyle. The component should follow and be tracked by the release policy of the container it was registered in, however understand this isn't really the desired behaviour for transients.

Unfortunately the fix for this won't be so easy, as you can see child containers haven't really been designed.

Windsor however has followed the rule that if you call Resolve you must also call Release even for transients, which would fix the unit test.

fir3pho3nixx:
Closing as per https://github.com/castleproject/Windsor/issues/338

PR #340, opened 2017-09-18
Obsolete logging facility enum

fir3pho3nixx:
That felt really good! Thanks for the leading PR ....


fir3pho3nixx:
Whoops too soon ...

jonorossi:
Not sure where this came from because I don't think we are ready for Windsor v5, however if you want to remove that much from the logging facility in v5 then we'll need other ways to actually configure the logger factory config file and log level because the logging facility currently creates the logger factory instance.

We'd have to change:
```
LogUsing<TLoggerFactory>(TLoggerFactory loggerFactory)
```

to:
```
LogUsing(ILoggerFactory loggerFactory)
```

Which means the logger factory can't automatically have dependencies resolved from the container. Obviously if this is the direction we'd want to take adding the new APIs now would be the way to go.

Would love to hear your thoughts here.

PR #341, opened 2017-09-18
Obsolete logging facility V5

fir3pho3nixx:
Again this felt great, thanks for leading the way ....


src/Castle.Windsor.Tests/LoggingFacility/Log4NetFacilityLognameOverrideTests.cs
```diff
@@ -47,15 +47,15 @@ public void Teardown()
...
-		[Test]
+		[Test] // Why override.? does not make sense, not what we told the logger to do.
 		public void OverrideTest()
 		{
 			container.Register(Component.For(typeof(SimpleLoggingComponent)).Named("component"));
 			var test = container.Resolve<SimpleLoggingComponent>("component");
 
 			test.DoSomething();
 
-			String expectedLogOutput = String.Format("[INFO ] [Override.{0}] - Hello world" + Environment.NewLine, typeof(SimpleLoggingComponent).FullName);
```
jonorossi: "Override" was the log name you removed above, this is the only purpose of this whole test class.

src/Castle.Facilities.Logging/LoggerImplementation.cs
```diff
@@ -1,41 +0,0 @@
...
-{
-	using System;
-
-	using Castle.Core.Logging;
-
-	/// <summary>
-	///   The supported <see cref = "ILoggerFactory" /> implementations.
-	/// </summary>
-	[Obsolete("A logger factory implementation type should be provided via LogUsing<T>(), this will be removed in the future.")]
-	public enum LoggerImplementation
```
jonorossi: I didn't really think we are ready to start on Windsor v5, otherwise I would have just done this rather than marking things obsolete which we haven't even released yet.

src/Castle.Facilities.Logging/LoggerResolver.cs
```diff
@@ -22,11 +22,6 @@ namespace Castle.Facilities.Logging
 	using Castle.MicroKernel;
 	using Castle.MicroKernel.Context;
 
-	/// <summary>
-	///   Custom resolver used by Windsor. It gives
-	///   us some contextual information that we use to set up a logging
-	///   before satisfying the dependency
-	/// </summary>
```
jonorossi: Why remove the comment?

src/Castle.Facilities.Logging/LoggingFacility.cs
```diff
@@ -12,403 +12,43 @@
...
-		public LoggingFacility LogUsing<TLoggerFactory>()
-			where TLoggerFactory : ILoggerFactory
-		{
-			loggerImplementation = LoggerImplementation.Custom;
-			loggingFactoryType = typeof(TLoggerFactory);
-			return this;
-		}
+		private ILoggerFactory loggerFactory;
 
 		public LoggingFacility LogUsing<TLoggerFactory>(TLoggerFactory loggerFactory)
```
jonorossi: No need for `TLoggerFactory` if you only end up using the `ILoggerFactory` instance.

src/Castle.Facilities.Logging/LoggingFacility.cs
```diff
@@ -12,403 +12,43 @@
...
 			where TLoggerFactory : ILoggerFactory
 		{
-			loggerImplementation = LoggerImplementation.Custom;
-			loggingFactoryType = typeof(TLoggerFactory);
 			this.loggerFactory = loggerFactory;
 			return this;
 		}
 
-		public LoggingFacility ConfiguredExternally()
```
jonorossi: If we are going to remove this, `WithConfig`, `WithLevel` and `WithAppConfig` then we should deprecate them in a v4.x release first.

src/Castle.Windsor.Tests/LoggingFacility/CustomFacilityTests.cs
```diff
@@ -1,58 +0,0 @@
...
-
-	using CastleTests.LoggingFacility.Tests.Classes;
-
-	using NUnit.Framework;
-
-	[TestFixture]
-	public class CustomFacilityTests
-	{
-		[Test]
-		public void ReadCustomFacilityConfigFromXML()
```
jonorossi: What is wrong with allowing people to configure the logging facility by XML? `loggingApi` is the only property listed in the changelog that is being deprecated as it is assumed to be "custom".

src/Castle.Facilities.Logging/LoggingFacility.cs
```diff
@@ -12,403 +12,43 @@
...
-		{
-			if (loggingApi == LoggerImplementation.Custom)
-			{
-				throw new FacilityException("To use custom logger use LogUsing<TLoggerFactory>() method.");
-			}
-			loggerImplementation = loggingApi;
-			return this;
-		}
-
-		public LoggingFacility LogUsing<TLoggerFactory>()
```
jonorossi: Removing this obviously means the logger factory can't automatically have dependencies resolved from the container. If this is the direction we want to take deprecating this API now would be the way to go, otherwise our documentation literally tells people to do something that we are removing.

src/Castle.Windsor.Tests/LoggingFacility/Log4NetFacilityLognameOverrideTests.cs
fir3pho3nixx: Ok I see it now, got confused between what the constructor of the factory was doing and the CreateConfiguredContainer extension.

src/Castle.Facilities.Logging/LoggerImplementation.cs
fir3pho3nixx: We can hold off, will chat to you once you have finished having dinner.

jonorossi: As discussed, we need to mark some more functionality obsolete in a v4 release but leave the removal until v5.

src/Castle.Windsor.Tests/LoggingFacility/CustomFacilityTests.cs
jonorossi: We also need to start a discussion issue about possibly removing XML configuration in v5, we should keep things consistent, either all the facilities should support it or none should.

src/Castle.Facilities.Logging/LoggerResolver.cs
fir3pho3nixx: Dont really like XML comments, they are sometimes not relevant or describe the mundanely obvious. It is leads to a yo-yo effect in the code file with regards to scrolling. Apart from giving you nice descriptions in the VS tooling I don't see much value in having them. Can put these back if you want, I don't feel strongly that they should be removed. Just makes reading code easier.

src/Castle.Facilities.Logging/LoggingFacility.cs
fir3pho3nixx: This was perhaps and eager cut, just that I saw the richness of the constructors for logger factories and thought this potentially hides features that are discoverable. Can bring this back but perhaps will refer back to the issue for further discussion before we accept this.

src/Castle.Facilities.Logging/LoggingFacility.cs
fir3pho3nixx: Correct.

src/Castle.Facilities.Logging/LoggingFacility.cs
fir3pho3nixx: Coming to conclusion that all of this is v5, they are all breaking changes.

src/Castle.Windsor.Tests/LoggingFacility/CustomFacilityTests.cs
fir3pho3nixx: Will refer back to issue on this and flesh this out with more discussion from our call.

fir3pho3nixx:
Adding the V5 milestone to this. I think I will bail out of this for now and start working on explicit assembly versions as we discussed. I am even happy to close this out once we have made more traction on the v4 stuff, dont like things hanging around open. I also have this on branch so it is really not a problem.

PR #342, opened 2017-09-19
Explicit Package Versions for Castle

fir3pho3nixx:
I tried a number of approaches, this was the cleanest way I could think of doing after backing out previous attempts.

image


tools/Explicit.NuGet.Versions/Explicit.NuGet.Versions.csproj
```diff
@@ -0,0 +1,15 @@
...
+	<PropertyGroup>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net45</TargetFramework>
+		<OutputPath>build\</OutputPath>
+		<AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+		<AssemblyName>nev</AssemblyName>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="DotNetZip" Version="1.10.1" />
```
jonorossi: I assume `System.IO.Compression` isn't up to the job for some reason?

tools/Explicit.NuGet.Versions/Program.cs
```diff
@@ -0,0 +1,95 @@
...
+			    nuspecXmlDocument.LoadXml(packageFile.Value.Contents);
+			    SetPackageDepencyVersionsToBeExplicitForXmlDocument(nuspecXmlDocument, dependencyNugetId);
+			    var updatedNuspecXml = nuspecXmlDocument.OuterXml;
+			    packageMetaData[packageFile.Key].Contents = updatedNuspecXml;
+		    }
+	    }
+
+	    private static void SetPackageDepencyVersionsToBeExplicitForXmlDocument(XmlDocument nuspecXmlDocument, string nugetIdFilter)
+	    {
+		    WalkDocumentNodes(nuspecXmlDocument.ChildNodes, node =>
```
jonorossi: Wow, I haven't seen System.Xml in new code for quite some time, rather than a single XPathSelectElements. It'll do the job though.

buildscripts/build.cmd
```diff
@@ -40,7 +41,9 @@ dotnet restore ./src/Castle.Windsor.Tests/Castle.Windsor.Tests.csproj
 GOTO build
 
 :build
+dotnet build ./tools/Explicit.NuGet.Versions/Explicit.NuGet.Versions.sln
 dotnet build Castle.Windsor.sln -c %Configuration%
+.\tools\Explicit.NuGet.Versions\build\nev.exe ".\build" "castle."
```
jonorossi: Won't this also incorrectly update Castle.Core?

tools/Explicit.NuGet.Versions/Explicit.NuGet.Versions.csproj
fir3pho3nixx: The ZipFile object expects you to extract the entire contents and repack it. This allows me to do it by doing an in-place swap of the nuspec file.

tools/Explicit.NuGet.Versions/Program.cs
fir3pho3nixx: Had issues with namespacing, rather than add a namespace table and use XPaths I though this would be more reillient to nuspec namespace changes.

buildscripts/build.cmd
fir3pho3nixx: Don't we want explicit versioning for all of it? Not a problem just need to add more commands to target specific package id's in the solution.

jonorossi: We've only discussed explicit versioning within a project, however I **don't** think Windsor should require a specific version of Castle Core assemblies.

fir3pho3nixx: All fixed

castle-windsor

castle-loggingfacility

PR #343, opened 2017-09-22
Wrong constructor being chosen for open generic service type registrations

### LINKED ISSUE #136: Wrong c'tor get's choosed
Accordingly to https://github.com/castleproject/Windsor/blob/master/docs/how-constructor-is-selected.md the constructor with the most satisfiable parameters gets picked.

In the following example, Windsor chooses the parameterless c'tor on MyControl although the c'tor with the parameter of type `IMyUtils<ArrayList>` can be resolved as proofed by `lResolvedFixedHelper`

Registering a generic type-unbound implementation for `IMyHelper<>` seems to fix the problem and the expected c'tor is called on MyControl.

```
using System;
using System.Collections;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using Castle.Windsor;

namespace ConsoleApplication15
{
    class Program
    {
        static void Main(string[] args)
        {
            var lContainer =  new WindsorContainer();
            var lGenericUtilsReg =
                Component.For(typeof(IMyUtils<>)).
                ImplementedBy(typeof(MyGenericUtilsImpl<>));

            lContainer.Register(lGenericUtilsReg);

            var lFixedHelperReg =
               Component.For(typeof(IMyHelper<ArrayList>)).
               ImplementedBy(typeof(MyArrayListFixedHelperImpl));

            lContainer.Register(lFixedHelperReg);

            var lControlReg =
               Component.For(typeof(MyControl)).
               ImplementedBy(typeof(MyControl));

            lContainer.Register(lControlReg);

            // Commenting the following code in, let's the example run as expected.
            //var lGenericHelperReg =
            //   Component.For(typeof(IMyHelper<>)).
            //   ImplementedBy(typeof(MyGenericHelperImpl<>));

            //lContainer.Register(lGenericHelperReg);

            var lResolvedFixedHelper = lContainer.Resolve<IMyHelper<ArrayList>>();
            if (lResolvedFixedHelper == null)
            {
                throw new ApplicationException("Should not happen.");
            }

            var lResolvedControl = lContainer.Resolve<MyControl>();
            if (lResolvedControl == null)
            {
                throw new ApplicationException("Should not happen.");
            }
        }
    }

    public class MyControl
    {
        public MyControl()
        {
            throw new ApplicationException("Should not be used.");
        }

        public MyControl(IMyUtils<ArrayList> aUtils)
        {
            Console.WriteLine("MyControl: ctor " + aUtils.GetType().Name);
        }
    }

    public class MyGenericUtilsImpl<T> : IMyUtils<T>
    {
        public MyGenericUtilsImpl(IMyHelper<T> aUtils)
        {
            Console.WriteLine("MyGenericUtilsImpl: ctor " + aUtils.GetType().Name);
        }
    }

    public class MyGenericHelperImpl<T> : IMyHelper<T>
    {
        public MyGenericHelperImpl()
        {
            Console.WriteLine("MyGenericHelperImpl: ctor");
        }
    }

    public class MyArrayListFixedHelperImpl : IMyHelper<ArrayList>
    {
        public MyArrayListFixedHelperImpl()
        {
            Console.WriteLine("MyArrayListFixedHelperImpl: ctor");
        }
    }

    public interface IMyUtils<T> { }

    public interface IMyHelper<T> { }
}
```

fir3pho3nixx:
This hopefully fixes #136.

### DIFF
--- a/src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
+++ b/src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
@@ -1,0 +1,64 @@
+﻿// Copyright 2004-2017 Castle Project - http://www.castleproject.org/
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace CastleTests.Windsor.Tests
+{
+	using System;
+	using System.Collections;
+
+	using Castle.MicroKernel.Registration;
+
+	using NUnit.Framework;
+
+	// Please see DefaultDependencyResolver -> HasAnyComponentInValidState -> RebuildOpenGenericHandlersWithClosedGenericSubHandlers
+	// 
+	// The problem here is that DefaultGenericHandler's will always be in an HandlerState.WaitingDependency state for open generic types.
+	// IOpenGeneric<> is registered in the container as an open generic handler. When resolving Control -> ctor(IOpenGeneric<ArrayList>) then the wrong
+	// constructor gets chosen for the Control class.  
+	// 
+	// In this case the DefaultGenericHandler of type IOpenGeneric<> should create a sub DefaultGenericHandler which is bound to IOpenGeneric<ArrayList>
+	// and submit whether it is resolvable or not during the DefaultDependencyResolver -> HasAnyComponentInValidState logic.
+
+	[TestFixture]
+	public class OpenGenericConstructorSelectionTestCase : AbstractContainerTestCase
+	{
+		[Test]
+		[Bug("https://github.com/castleproject/Windsor/issues/136")]
+		public void Resolves_using_most_greedy_constructor_when_having_open_generic_container_registrations_with_inferred_generic_parameters()
+		{
+			Container.Register(Component.For(typeof(IOpenGeneric<>)).ImplementedBy(typeof(OpenGeneric<>)));
+			Container.Register(Component.For(typeof(IClosedGenericArrayList<ArrayList>)).ImplementedBy(typeof(ClosedGenericArrayList)));
+			Container.Register(Component.For(typeof(Control)));
+			Container.Resolve<Control>();
+		}
+
+		public class Control
+		{
+			// Least Greedy has no parameters, should NOT be chosen.
+			public Control() { throw new Exception("The default constructor should not be chosen!"); }
+
+			// Most greedy has parameter, should be chosen.
+			public Control(IOpenGeneric<ArrayList> param) { }
+		}
+
+		public class OpenGeneric<T> : IOpenGeneric<T>
+		{
+			public OpenGeneric(IClosedGenericArrayList<T> param) { }
+		}
+
+		public class ClosedGenericArrayList : IClosedGenericArrayList<ArrayList> { }
+		public interface IOpenGeneric<T> { }
+		public interface IClosedGenericArrayList<T> { }
+	}
+}
--- a/src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs
+++ b/src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs
@@ -280,6 +280,12 @@ namespace Castle.MicroKernel.Handlers
 			}
 		}
 
+		public IHandler ConvertToClosedGenericHandler(Type service, CreationContext openGenericContext)
+		{
+			var closedType = this.GetClosedImplementationType(openGenericContext, false);
+			return this.GetSubHandler(closedType, service);
+		}
+
 		private Type GetClosedImplementationType(CreationContext context, bool instanceRequired)
 		{
 			if (ComponentModel.Implementation == typeof(LateBoundComponent))
--- a/src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
+++ b/src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
@@ -265,8 +265,26 @@ namespace Castle.MicroKernel.Resolvers
 			}
 
 			var handlers = kernel.GetHandlers(service);
-			return handlers.Where(handler => handler.IsBeingResolvedInContext(context) == false)
-				.Any(IsHandlerInValidState);
+			var nonResolvingHandlers = handlers.Where(handler => handler.IsBeingResolvedInContext(context) == false).ToList();
+			RebuildOpenGenericHandlersWithClosedGenericSubHandlers(service, context, nonResolvingHandlers);
+			return nonResolvingHandlers.Any(handler => IsHandlerInValidState(handler));
+		}
+
+		private void RebuildOpenGenericHandlersWithClosedGenericSubHandlers(Type service, CreationContext context, List<IHandler> nonResolvingHandlers)
+		{
+			if (context.RequestedType != null && service.GetTypeInfo().IsGenericType)
+			{
+				// Remove DefaultGenericHandlers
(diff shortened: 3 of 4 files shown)


src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		public class OpenGeneric<T> : IOpenGeneric<T>
+		{
+			public OpenGeneric(IClosedGenericArrayList<T> param) { }
+		}
+
+		public class ClosedGenericArrayList : IClosedGenericArrayList<ArrayList> { }
+		public interface IOpenGeneric<T> { }
+		public interface IClosedGenericArrayList<T> { }
```
jonorossi: I don't see a need for two interfaces here, one for open and another for closed? I think the test would read better if it was a single interface so your mind isn't trying to work out what the difference between the two are, when the difference is just how they are used.

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
```diff
@@ -0,0 +1,67 @@
...
+		{
+			Container.Register(Component.For(typeof(IOpenGeneric<>)).ImplementedBy(typeof(OpenGeneric<>)));
+			Container.Register(Component.For(typeof(IClosedGenericArrayList<ArrayList>)).ImplementedBy(typeof(ClosedGenericArrayList)));
+			Container.Register(Component.For(typeof(Control)).ImplementedBy(typeof(Control)));
+			Container.Resolve<Control>();
+		}
+
+		public class Control
+		{
+			//[DoNotSelect] // Least Greedy has no parameters, should NOT be chosen.
```
jonorossi: I wouldn't mention `[DoNotSelect]` here in the comments because this really has nothing to do with this problem, `DoNotSelectAttribute` skips the constructor completely where this is about selection. I assume you just accidentally left it from developing.

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
```diff
@@ -0,0 +1,67 @@
...
+	[TestFixture]
+	public class OpenGenericConstructorSelectionTestCase : AbstractContainerTestCase
+	{
+		[Test]
+		[Bug("https://github.com/castleproject/Windsor/issues/136")]
+		public void Resolves_using_most_greedy_constructor_when_having_open_generic_container_registrations_with_inferred_generic_parameters()
+		{
+			Container.Register(Component.For(typeof(IOpenGeneric<>)).ImplementedBy(typeof(OpenGeneric<>)));
+			Container.Register(Component.For(typeof(IClosedGenericArrayList<ArrayList>)).ImplementedBy(typeof(ClosedGenericArrayList)));
+			Container.Register(Component.For(typeof(Control)).ImplementedBy(typeof(Control)));
```
jonorossi: No need for `ImplementedBy`.

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
```diff
@@ -0,0 +1,67 @@
...
+		Please see DefaultDependencyResolver -> HasAnyComponentInValidState -> RebuildOpenGenericHandlersWithClosedGenericSubHandlers
+
+		The problem here is that DefaultGenericHandler's will always be in an HandlerState.WaitingDependency state for open generic types.
+		IOpenGeneric<> is registered in the container as an open generic handler. When resolving Control -> ctor(IOpenGeneric<ArrayList>) then the wrong
+		constructor gets chosen for the Control class.  
+
+		In this case the DefaultGenericHandler of type IOpenGeneric<> should create a sub DefaultGenericHandler which is bound to IOpenGeneric<ArrayList>
+		and submit whether it is resolvable or not during the DefaultDependencyResolver -> HasAnyComponentInValidState logic.
+		
+	*/
```
jonorossi: I doubt it is in our style guide however I dislike multiline comments because their indentation always gets messy, I've prefer this change to single line comments like the license header.

src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs
```diff
@@ -280,6 +280,12 @@ private void CloneParentProperties(ComponentModel newModel)
 			}
 		}
 
+		public IHandler ConvertToClosedGenericHandler(Type service, CreationContext openGenericContext)
+		{
+			var closedType = this.GetClosedImplementationType(openGenericContext, true); // Not sure about required
```
jonorossi: `instanceRequired` seems to basically determine whether `DefaultGenericHandler` will throw (`true`) or return null (`false`). Since you are using this from `CanResolve` you probably want `false`.

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
```diff
@@ -265,8 +265,26 @@ private bool HasAnyComponentInValidState(Type service, DependencyModel dependenc
 			}
 
 			var handlers = kernel.GetHandlers(service);
-			return handlers.Where(handler => handler.IsBeingResolvedInContext(context) == false)
-				.Any(IsHandlerInValidState);
+			var nonResolvingHandlers = handlers.Where(handler => handler.IsBeingResolvedInContext(context) == false).ToList();
+			RebuildOpenGenericHandlersWithClosedGenericSubHandlers(service, context, nonResolvingHandlers);
```
jonorossi: Do you think this change will have any impact on the `IGenericImplementationMatchingStrategy` extension point? Not sure how many unit tests that has.

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
fir3pho3nixx: All done

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
fir3pho3nixx: Copy paste issue, fixed

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
fir3pho3nixx: I was investigating options around this problem first, never knew donotselect was a thing! :) Removed.

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
fir3pho3nixx: No neither did I, it was only when the resolution chain went via an open generic that the wrong ctor got chosen. I originally hammered a test suite out with about 15 different tests which I could not get to fail thinking along exactly the same lines you were. It was only once I replicated and cleaned this up from the OP I could get it to fail. 

I can offer to rename some stuff but I would need your input here. This is why I added the long comment at the start of the test.

src/Castle.Windsor/MicroKernel/Handlers/DefaultGenericHandler.cs
fir3pho3nixx: Thanks for this. It is very very important to note that this logic is now used from both can resolve and resolve paths. The fix however was really in the default sub resolver during the resolve process. It was a bastard and took a full day of debugging to work it out.

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
fir3pho3nixx: Hope not or else GenericImplementationWithGreaterArityThanServiceTestCase && GenericMultipleServicesComponentResolutionTestCase should be chucking up red flags. 

This is covered by tests that assert that these bits still work.

src/Castle.Windsor.Tests/Windsor.Tests/OpenGenericConstructorSelectionTestCase.cs
jonorossi: Ah yep, I get it now, the big comment block makes sense, subtle detail!

PR #344, opened 2017-09-23
Fixed typed factory out of order disposal

### LINKED ISSUE #114: TypedFactory Disposes Before Component That Depends on it
If a typed factory is a dependency of a disposable component, it is already disposed by the time dispose is called on that component. 

If I directly implement the factory myself rather than using the typed factory, everything behaves as I would expect.

[This SO Question also details the issue](http://stackoverflow.com/questions/34361165/typedfactory-disposes-before-component-using-it-as-a-dependency)

I have attached a test solution below that shows this behavior (although it is not actually useful itself).

[WindsorTypedFactoryDisposalTest.zip](https://github.com/castleproject/Windsor/files/67637/WindsorTypedFactoryDisposalTest.zip)

jnm2:
Fixes #114. See implementation rationale at https://github.com/castleproject/Windsor/issues/114#issuecomment-331667950.

This has been causing no small amount of problems for my company, so hopefully this fix and rationale are adequate.

/cc @fir3pho3nixx @jonorossi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Bugfixes:
 - Fix binding errors because assembly version had too much detail, assembly version is now x.0.0.0 (@fir3pho3nixx, #329)
 - Update Castle Core to 4.1.1 (TODO: update again) to resolve assembly version problems because Castle Core also had too much detail
 - Explicit package versioning applied within solution to avoid maligned NuGet upgrades for lock step versioned packages. (@fir3pho3nixx, Core/#292)
+- Fixed typed factory out of order disposal (@jnm2, #344)
 
 Deprecations:
 - Logging Facility's `LoggerImplementation` enum, `UseLog4Net` and `UseNLog` methods are deprecated in favour of `LogUsing<T>`, this includes the `loggingApi` property for XML configuration (@jonorossi, #327)
--- a/src/Castle.Windsor.Tests/Facilities/TypedFactory/TypedFactoryDisposeOrderTestCase.cs
+++ b/src/Castle.Windsor.Tests/Facilities/TypedFactory/TypedFactoryDisposeOrderTestCase.cs
@@ -1,0 +1,74 @@
+﻿// Copyright 2004-2017 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


hammett:
LGTM. I think you deserve some kind of medal for this one, even though the fix is simple, getting there was a brave man's endeavour. Thanks!

jnm2:
@jonorossi Done. The next thing on my mind is, how soon would you consider a hotfix release to NuGet? If worst comes to worst I can build a package myself (your CI artifacts have versions like `0.0.162` which won't be seen as newer than `4.0.0`) but I'd rather not have to worry about that.

fir3pho3nixx:
@jnm2 - The build version if defaulted to zero [here](https://github.com/castleproject/Windsor/blob/master/buildscripts/common.props#L7). What you are seeing is a noddy appveyor counter running in the background which has no relevance to the release. Dev builds only. 

When I worked on porting this project to AppVeyor @jonorossi and I made sure that tag versions for releases flow through to the final published NuGets. Tag = deploy.

If we are following semver than this looks like it will be in a 4.0.1 release.

jnm2:
4.0.1 sounds good. Would you be willing to put at least a prerelease of 4.0.1 on NuGet within a week or so?

fir3pho3nixx:
Faster than that fella! That is why I reworked everything to use appveyor. We have no excuses for releasing early and often anymore :)

Before AppVeyor we used to hoard everything in master with no visibility of the background TeamCity server when merging PR's. 

Your fix is shall we say ... imminent.

jonorossi:
Merged.

@jnm2 @fir3pho3nixx definitely, just need to cut a fresh Castle Core release (https://github.com/castleproject/Core/issues/306) and get Windsor updated to end those assembly binding version problems.

Logged an issue so we can get this version out: https://github.com/castleproject/Windsor/issues/347

MishaD87:
Hello,
I am currently having described problem. I have Castle.Windsor 4.1.1 and Castle.Core 4.3.1.
I have transient module that implements IDisposible and has dependency on typed factory.
Factory creates instances of objects with Singleton life cycle. 
I create objects with factory in my module and trying to release them in Dispose method.
But getting this message "The factory was disposed and can no longer be used"

jnm2:
@MishaD87 Would you be willing to open a new issue and provide the smallest sample code you can which demonstrates this? That'll take a lot of the guesswork out.

PR #352, opened 2017-10-09
Harmonise Castle Core and Windsor

fir3pho3nixx:
Adding missing xml documentation to NuGets and including the documentation in the solution

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@ _ReSharper*
 *.cache
 * Thumbs.db
 
-#leftovers from merge
+# Leftovers from merging
 *.orig
 *.bak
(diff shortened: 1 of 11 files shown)


jonorossi:
[wrote similar comment in Castle Core PR]

Windsor 4.1.0 already has XML files for each target framework, the new .NET tooling must automatically generate them, however not sure what is different with Castle Core which also didn't specify the DocumentationFile element.

fir3pho3nixx:
Yip we can close this.

fir3pho3nixx:
I did bring the docs into the solution in the other PR anyway. I will double check those also.

PR #353, opened 2017-10-11
fixed bug in RequireScope

WahidBitar:
the RequireScope method was returning null if ObtainCurrentScope returns a value which is very strange

### DIFF
--- a/src/Castle.Windsor/MicroKernel/Lifestyle/LifestyleExtensions.cs
+++ b/src/Castle.Windsor/MicroKernel/Lifestyle/LifestyleExtensions.cs
@@ -31,12 +31,7 @@ namespace Castle.MicroKernel.Lifestyle
 
 		public static IDisposable RequireScope(this IKernel kernel)
 		{
-			var current = Scope.ObtainCurrentScope();
-			if (current == null)
-			{
-				return kernel.BeginScope();
-			}
-			return null;
+			return Scope.ObtainCurrentScope() ?? kernel.BeginScope();
 		}
 
 		public static IDisposable BeginScope(this IWindsorContainer container)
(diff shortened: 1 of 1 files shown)


fir3pho3nixx:
Looks like some tests are failing. Could you also consider adding a test case for this so we can see the problem?

WahidBitar:
OK sure I'll try soon

WahidBitar:
I found the failed Test.

```
[Test]
public void Requiring_scope_within_parent_scope_uses_parent_scope()
{
   Container.Register(Component.For<A>().LifeStyle.Scoped());
   using (Container.BeginScope())
   {
	var a = Container.Resolve<A>();
	using (var scope = Container.RequireScope())
	{
	   var aa = Container.Resolve<A>();
	   Assert.AreSame(a, aa);
	   Assert.IsNull(scope);
	}
   }
}
```

So for me, it doesn't make sense to return null scope if you found the parent scope.
**I assumed it should return the parent scope or create new one**, in this case, we should check that the scope is not null instead of checking that it should be null `Assert.IsNull(scope);`

jonorossi:
> So for me, it doesn't make sense to return null scope if you found the parent scope.
> I assumed it should return the parent scope or create new one, in this case, we should check that the scope is not null instead of checking that it should be null Assert.IsNull(scope);

The problem with returning the parent scope is that you use it in a `using` block, which will then dispose the parent scope prematurely.

We at least need to add some XML comments to those methods to make this clear.

WahidBitar:
> The problem with returning the parent scope is that you use it in a using block, which will then dispose the parent scope prematurely.

Yes, you're right. Sorry, this goal wasn't clear for me at the first look.


> We at least need to add some XML comments to those methods to make this clear.

It would be great.

Thank you for the speed correction

jonorossi:
> Yes, you're right. Sorry, this goal wasn't clear for me at the first look.

No problem. The other thing to note is that these methods return `IDisposable` not `ILifetimeScope` explicitly so they don't get used to fetch the scope. I've added some XML docs in #354.

fir3pho3nixx:
Just approved the PR.

PR #354, opened 2017-10-13
Add XML docs to scoping extensions and tidy up CallContextLifetimeScope

jonorossi:
Added docs as discussed in #353.

### DIFF
--- a/src/Castle.Windsor/MicroKernel/Lifestyle/Scoped/CallContextLifetimeScope.cs
+++ b/src/Castle.Windsor/MicroKernel/Lifestyle/Scoped/CallContextLifetimeScope.cs
@@ -1,4 +1,4 @@
-// Copyright 2004-2014 Castle Project - http://www.castleproject.org/
+// Copyright 2004-2017 Castle Project - http://www.castleproject.org/
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


jonorossi:
I've removed two public constructors from `CallContextLifetimeScope` because they do nothing different than the default constructor. I don't see any reason this should be classified as a breaking change because no one should be using this class directly even though it is public.

/cc @fir3pho3nixx

src/Castle.Windsor/MicroKernel/Lifestyle/Scoped/CallContextLifetimeScope.cs
```diff
@@ -22,81 +22,73 @@ namespace Castle.MicroKernel.Lifestyle.Scoped
...
 				if (parentScope != null)
 				{
 					SetCurrentScope(parentScope);
 				}
-#if FEATURE_REMOTING
 				else
 				{
-					CallContext.FreeNamedDataSlot(keyInCallContext);
-				}
+#if FEATURE_REMOTING
```
fir3pho3nixx: Interesting, why would you let the else potion of this evaluate?

jonorossi: It isn't, I just moved the `#ifdef` inside the `else` statement so it just wraps the body.

I was thinking about clearing the `AsyncLocal`'s value then realised that is pointless, and wondered why we were clearing the value from the `CallContext`. Apparently logical call contexts in .NET 4.5+ are now copy-on-write, so we shouldn't need to call `FreeNamedDataSlot` anymore because the entry should disappear as you exit the method that added it. I deferred making the change since it isn't documented on MSDN.

Issue #325, opened 2017-09-11
IDisposable and UsingFactoryMethod


aaron-hammond:
Consider: 

```
public class ServiceInstaller : IWindsorInstaller
    {
        public void Install(IWindsorContainer container, IConfigurationStore store)
        {
            container.Register(
                Component.For<ISomeConnectionService>()
                    .ImplementedBy<SomeConnectionService>()
                    .LifestyleTransient()
                    .IsDefault());
            
            container.Register(
                Component.For<FailoverDatabaseConnectionExecutor>()
                .ImplementedBy<FailoverDatabaseConnectionExecutor>()
                .LifestyleTransient());

            container.Register(Component.For<IDatabaseConnectionExecutor>()
                .UsingFactoryMethod(CreateDatabaseConnectionExecutor)
                .LifestyleTransient()
                .IsDefault());
        }

        private static IDatabaseConnectionExecutor CreateDatabaseConnectionExecutor(IKernel kernel)
        {
            return kernel.Resolve<FailoverDatabaseConnectionExecutor>();
        }
    }
    
    public interface IDatabaseConnectionExecutor
    {
    }

    public class SomeConnectionService : ISomeConnectionService, System.IDisposable
    {
        public SomeConnectionService()
        {

        }

        public void Dispose()
        {
        }
    }

    public interface ISomeConnectionService
    {
    }

    public class FailoverDatabaseConnectionExecutor : IDatabaseConnectionExecutor
    {
        private readonly ISomeConnectionService _someConnectionService;

        public FailoverDatabaseConnectionExecutor(ISomeConnectionService someConnectionService)
        {
            _someConnectionService = someConnectionService;
        }
    }
```

When attempting to inject ISomeConnectionService i receive the following error:

> Instance FailoverDatabaseConnectionExecutor of component Late bound IDatabaseConnectionExecutor is already being tracked. The factory method providing instances of the component is reusing instances, but the lifestyle of the component is Transient which requires new instance each time. In most cases it is advised for the factory method not to be handling reuse of the instances, but to chose a lifestyle that does that appropriately. Alternatively, if you do not wish for Windsor to track the objects coming from the factory change your regustration to '.UsingFactoryMethod(yourFactory, managedExternally: true)'

[...]

PR #373, opened 2018-01-12
Burdens are not Release Policy aware from TypedFactories for disposables

fir3pho3nixx:
After investigating this issue extensively: https://github.com/castleproject/Windsor/issues/325

Going Dan North on this.

Given: 

Type `Foo` implements IDisposable.
Type `Foo` is registered via a Typed Factory using a service type with an interface `IFoo`.
Type `Foo` is registered outside of a typed factory with the implementing service type `Foo`.
Type `Foo` has a life style of transient for both typed facotry and non typed factory registrations.
Type `Foo` is a late bound component because of the single initial typed factory registration.
Type `Foo` has IsDefault status via the Typed Factory registration.
The typed factory method for Foo resolves using the windsor microkernel using the delegate component activator.

When:

I resolve IFoo.

Then:

I should have only one instance. 
I should only track instances once via the release policy for dispoables.
I should have two burdens from both the typed factory and the non typed factory registration. 
I should not have to track sealed non disposables at the parent level that consumes disposable dependencies.
I do care about dependent disposables for top level graph non disposables.

### DIFF
--- a/src/Castle.Windsor.Tests/Registration/UsingFactoryMethodTestCase.cs
+++ b/src/Castle.Windsor.Tests/Registration/UsingFactoryMethodTestCase.cs
@@ -380,7 +380,8 @@ namespace CastleTests.Registration
 
 			var component = Kernel.Resolve<SealedComponentWithDependency>();
 
-			Assert.IsTrue(Kernel.ReleasePolicy.HasTrack(component));
+			// Why are we tracking sealed late bound non disposables? This is bonkers. Surely we are only interested in disposable dependencies?
+			// Assert.IsTrue(Kernel.ReleasePolicy.HasTrack(component));
 			Assert.IsTrue(Kernel.ReleasePolicy.HasTrack(component.Dependency));
 		}
 
--- a/src/Castle.Windsor/MicroKernel/Burden.cs
+++ b/src/Castle.Windsor/MicroKernel/Burden.cs
@@ -25,13 +25,15 @@ namespace Castle.MicroKernel
 	public class Burden
 	{
 		private readonly IHandler handler;
+		private readonly IReleasePolicy releasePolicy;
 		private Decommission decommission = Decommission.No;
 
 		private List<Burden> dependencies;
 
-		internal Burden(IHandler handler, bool requiresDecommission, bool trackedExternally)
+		internal Burden(IHandler handler, IReleasePolicy releasePolicy, bool requiresDecommission, bool trackedExternally)
 		{
 			this.handler = handler;
+			this.releasePolicy = releasePolicy;
 			TrackedExternally = trackedExternally;
 			if (requiresDecommission)
 			{
@@ -89,7 +91,32 @@ namespace Castle.MicroKernel
 		/// </summary>
 		public bool RequiresPolicyRelease
 		{
-			get { return TrackedExternally == false && RequiresDecommission; }
+			get
+			{
+				var isAlreadyTracked = releasePolicy.HasTrack(Instance);
+				var isLateBound = Model.Implementation == typeof(LateBoundComponent);
+
+				// If we track externally, we want to bail out early
+				if (TrackedExternally)
+				{
+					return false;
+				}
+
+				// If we are not tracked extenerally but we are not late bound go ahead with default resolve
+				if (!isLateBound && !TrackedExternally)
+				{
+					return RequiresDecommission;
+				}
+
+				// We are not tracked externally, but we are late bound, if we have disposables we are burden N+1 tracked via TypedFactory and non TypeFactory component registrations for any given component type <T>
+				if (isLateBound && !TrackedExternally)
+				{
+					return RequiresDecommission && !isAlreadyTracked;
+				}
+
+				// Fallback logic for non type factory types T 
+				return !TrackedExternally && RequiresDecommission;
+			}
 		}
 
 		public bool TrackedExternally { get; set; }
--- a/src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
+++ b/src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
@@ -232,7 +232,7 @@ namespace Castle.MicroKernel.Context
 		public ResolutionContext EnterResolutionContext(IHandler handlerBeingResolved, bool trackContext,
 		                                                bool requiresDecommission)
 		{
-			var resolutionContext = new ResolutionContext(this, handlerBeingResolved, requiresDecommission, trackContext);
+			var resolutionContext = new ResolutionContext(this, handlerBeingResolved, ReleasePolicy, requiresDecommission, trackContext);
 			handlerStack.Push(handlerBeingResolved);
 			if (trackContext)
 			{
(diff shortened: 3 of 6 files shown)


fir3pho3nixx:
@jnm2 - Could you also please review this PR?

fir3pho3nixx:
@aaron-hammond - Could you also please review this PR?

src/Castle.Windsor/MicroKernel/Burden.cs
```diff
@@ -148,6 +175,7 @@ public void SetRootInstance(object instance)
 			Instance = instance;
 			if (decommission == Decommission.LateBound)
 			{
+				// BOOM!
```
fir3pho3nixx: Smoking gun ...

src/Castle.Windsor/MicroKernel/Burden.cs
```diff
@@ -89,7 +91,32 @@ public bool RequiresDecommission
 		/// </summary>
 		public bool RequiresPolicyRelease
 		{
-			get { return TrackedExternally == false && RequiresDecommission; }
+			get
```
fir3pho3nixx: Fucking A, we probably need new abstraction.

src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
```diff
@@ -232,7 +232,7 @@ public ResolutionContext EnterResolutionContext(IHandler handlerBeingResolved, b
 		public ResolutionContext EnterResolutionContext(IHandler handlerBeingResolved, bool trackContext,
 		                                                bool requiresDecommission)
 		{
-			var resolutionContext = new ResolutionContext(this, handlerBeingResolved, requiresDecommission, trackContext);
+			var resolutionContext = new ResolutionContext(this, handlerBeingResolved, ReleasePolicy, requiresDecommission, trackContext);
```
fir3pho3nixx: How did we miss this from 2011

src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
```diff
@@ -450,17 +450,19 @@ public class ResolutionContext : IDisposable
...
 			private readonly CreationContext context;
 			private readonly IHandler handler;
+			private readonly IReleasePolicy releasePolicy;
 			private readonly bool requiresDecommission;
 			private readonly bool trackContext;
 			private Burden burden;
 			private IDictionary extendedProperties;
 
-			public ResolutionContext(CreationContext context, IHandler handler, bool requiresDecommission, bool trackContext)
+			public ResolutionContext(CreationContext context, IHandler handler, IReleasePolicy releasePolicy, bool requiresDecommission, bool trackContext)
```
fir3pho3nixx: Again so needed, how did we miss this?

src/Castle.Windsor.Tests/Registration/UsingFactoryMethodTestCase.cs
```diff
@@ -380,7 +380,8 @@ public void Factory_created_sealed_non_disposable_services_with_factory_created_
 
 			var component = Kernel.Resolve<SealedComponentWithDependency>();
 
-			Assert.IsTrue(Kernel.ReleasePolicy.HasTrack(component));
+			// Why are we tracking sealed late bound non disposables? This is bonkers. Surely we are only interested in disposable dependencies?
+			// Assert.IsTrue(Kernel.ReleasePolicy.HasTrack(component));
```
fir3pho3nixx: Jesus christ! Why?

jnm2:
(Fyi the commit message has two typos, 'facotry' and 'dispoables.')

I don't sense anything wrong about this, though I'm aware that the meat of it may be going over my head a bit. I'd need someone to break down for me the significance of  `ReleasePolicy_tracking_the_same_instance_ twice_with_transient_lifestyle_and_factory_method_suggests_different_lifestyle`.

> Why are we tracking sealed late bound non disposables? This is bonkers. Surely we are only interested in disposable dependencies?

I'm not sure what 'sealed' means in this context. Do non-disposable instances need to be tracked so that their potentially disposable dependencies can be released at the appropriate time?

Another question: is it possible to configure an ad-hoc decommission handler on a non-disposable component?

> If we track externally, we want to bail out early

Can you help me understand what it means to be tracked externally and how we know for sure that whoever is doing the external tracking has all the necessary information?

fir3pho3nixx:
> I don't sense anything wrong about this, though I'm aware that the meat of it may be going over my head a bit. I'd need someone to break down for me the significance of ReleasePolicy_tracking_the_same_ instance_twice_with_transient_lifestyle_and_factory_method_suggests_different_lifestyle

This was probably added to help guard against misconfigured lifestyles when helpful exceptions were added. I am not sure it is valid anymore. My reasoning is this:

Factories are late bound components, from what I can tell they need burdens if they are to be tracked. If factories engage the vanilla resolve mechanism from the MicroKernel inside the factory having 2 burdens is unavoidable but you might end up with the same instance for a transient lifestyle. By extending the burdens to be release policy aware and using HasTrack functionality you can allow this to pass for the same resolution context. Maybe this helpful exception was masking over a potential bug? 

> I'm not sure what 'sealed' means in this context. Do non-disposable instances need to be tracked so that their potentially disposable dependencies can be released at the appropriate time?

You got me there, I will double check for memory leaks with transients. It seems reasonable to think this.

> Can you help me understand what it means to be tracked externally and how we know for sure that whoever is doing the external tracking has all the necessary information?

My understanding is when you ask components to be tracked externally you are telling windsor not to wrap them in burdens. The side effect of this if you have to release and dispose yourself.

jnm2:
Great, thanks for the explanations!

> If factories engage the vanilla resolve mechanism from the MicroKernel inside the factory having 2 burdens is unavoidable but you might end up with the same instance for a transient lifestyle.

Sorry to make you explain so much! I'm with you this far, but why is it that engaging the vanilla resolve mechanism results in two burdens? And how is it possible for you to end up with the same instance for a transient lifestyle?

fir3pho3nixx:
> Sorry to make you explain so much! 

No problem :)

> And how is it possible for you to end up with the same instance for a transient lifestyle?

If you have a `Foo` which takes `Bar` into it's constructor and `Bar` implements IDisposable. 

1. Register Bar as transient. 
2. Register Foo as transient
3. Register IFoo using factory which calls kernel.resolve<Foo> with IsDefault
4. Then call container.resolve<IFoo>

Because the factory calls kernel.resolve inside of it, it is effectively passing back the same transient from registration (2) in (3). Bizarrely you can do this without problems when you remove IDisposable from `Bar`.

> I'm with you this far, but why is it that engaging the vanilla resolve mechanism results in two burdens? 

Looks like Windsor is trying to track Burdens for `Bar` here during the resolution of Foo and finally resolution of IFoo. Because they are the same instance(because of calling resolve inside the factory method), Burdens have trouble tracking this twice. This is what led me to believe that they need to be release policy aware to be able to make use of the HasTrack feature. This was Bar will still get resolved whether or not it implements disposable. 

When doing that that though, a pretty important looking helpful exception no longer get's thrown :)

fir3pho3nixx:
Can re-open if we feel necessary.

jnm2:
Hopefully Krzysztof has some insight. Let me know if you think I might be able to help.

fir3pho3nixx:
@jnm2 - Thanks, I deleted the comment after I posted it. Happy to take this forward with @jonorossi with detailed discussions. 

@jonorossi might be out for a while due to personal reasons.

Issue #243, opened 2017-05-18
Use the latest NUnitTestAdapter for running tests (3.8.0-alpha1)


alinapopa:
Ref: #234 

The latest adapter does not run on both windows and linux. We can only progress this once the issue below is dealt with. 

https://github.com/nunit/nunit3-vs-adapter/issues/324

fir3pho3nixx:
Turns out this is a problem with nunit/xunit also. Created a repro @ https://github.com/fir3pho3nixx/nunit-adapter-vs-xunit which proves the case.

fir3pho3nixx:
@all - I have just seen the commit go in that fixes this in the test adapter. As soon as they release I will raise a PR

fir3pho3nixx:
[ping back from ms](https://github.com/Microsoft/vstest/issues/865#issuecomment-312255464)

fir3pho3nixx:
Just an update, there appears to be more problems with the latest nightly build that is in preview. Things appear to be going backwards. I am still actively investigating this ... will keep this issue updated with any pertinent info.

fir3pho3nixx:
This is definitely possible now with [netcoreapp2.0 now](https://github.com/castleproject/Windsor/pull/376/commits/bc2d2cb496a09a5c8c5f1a636bb44642a6a2c117).

fir3pho3nixx:
Can I go ahead an upgrade?

jonorossi:
> Can I go ahead an upgrade our test libraries + appveyor build?

Yep, go for it.

fir3pho3nixx:
Closing in anticipation of: https://github.com/castleproject/Windsor/pull/381

Issue #120, opened 2016-02-06
Integration for Microsoft.Extensions.DependencyInjection (ASP.NET Core)


cwe1ss:
Hi,
I created an integration for Microsoft.Extensions.DependencyInjection. 

You can find the code here: https://gist.github.com/cwe1ss/050a531e2711f5b62ab0

The code is based on the following links:
- https://github.com/autofac/Autofac/tree/master/src/Autofac.Extensions.DependencyInjection
- https://davidzych.com/asp-net-vnext-dependency-injection-with-castle-windsor/
- https://github.com/aspnet/Home/issues/263
- https://github.com/davezych/DependencyInjection/tree/windsor/src/Microsoft.Framework.DependencyInjection.Windsor
- http://stackoverflow.com/questions/28353187/how-can-i-get-started-with-asp-net-5-and-castle-windsor-for-dependency-injection

I adjusted the code for RC1 and fixed some issues:
- Components were added with .OnlyNewServices() which resulted in services that are supposed to exist multiple times not being added
- GetService<IEnumerable<TType>>() was not resolved properly. I had to add special handling for IEnumerable<>
- ASP.NET Core doesn't expect GetService() to return an exception if the component wasn't registered. (they use it for optional dependencies). For this reason, I added a check to Kernel.HasComponent()

I'm not sure why the call to BeginScope() in Startup.cs is necessary - if it's missing, the first request fails. Seems like ASP.NET Core doesn't start the scope for the first request or some Scoped service is requested before it does.

best regards!

cwe1ss:
I moved the code into a repository. [you can find it here](https://github.com/c3-ls/CastleWindsor-MsExtensionsIntegration).

jonorossi:
@cwe1ss sorry for not replying to your original post, I've had some personal things going on over the last couple of months which have really pulled me away from Castle.

Firstly thanks for building the integration. Unfortunately I've done zero with ASP.NET Core and the new DI framework built into it so can't really comment as to the problems you've encountered.

I assume you'd like Windsor to ship with this integration? You pasted links to code and problems, followed by "best regards", unfortunately it isn't going to get into Windsor without someone contributing the work. Your gist doesn't have a license, while your repo is licensed under the MIT license, contributions to Windsor must be licensed under the Apache 2 license.

[...]

Issue #328, opened 2017-09-11
Please provide a method like ValidateConfiguration() for tests


jnm2:
There's nothing in the documentation about validating a container. I want a guarantee under test that for all the dependencies the container knows about, you would not receive a configuration error.

This is the [best thing I can find](https://stackoverflow.com/a/8970760), courtesy of @mikehadlow:

```cs
private static void CheckForPotentiallyMisconfiguredComponents(IWindsorContainer container)
{
    var host = (IDiagnosticsHost)container.Kernel.GetSubSystem(SubSystemConstants.DiagnosticsKey);
    var diagnostics = host.GetDiagnostic<IPotentiallyMisconfiguredComponentsDiagnostic>();

    var handlers = diagnostics.Inspect();

    if (handlers.Any())
    {
        var message = new StringBuilder();
        var inspector = new DependencyInspector(message);

        foreach (IExposeDependencyInfo handler in handlers)
        {
            handler.ObtainDependencyDetails(inspector);
        }

        throw new MisconfiguredComponentException(message.ToString());
    }
}
```

There's nothing obvious about this code. If it is indeed the code you're recommend, please ship the code as a `ValidateConfiguration()`.

I'm surprised that such a test isn't recommended practice for every user. This is something I'd expect to find conspicuously linked from both the [readme](https://github.com/castleproject/Windsor/blob/master/docs/README.md) and the [FAQ](https://github.com/castleproject/Windsor/blob/master/docs/faq.md).

This is the closest thing I could find. It talks about testing various aspects of certain registrations but does not mention validating registrations to check specifically that all dependencies can be satisfied:  https://github.com/castleproject/Windsor/blob/master/docs/mvc-tutorial-part-3a-testing-your-first-installer.md

jnm2:
And, the above code is no good if you have with typed factories. So I'm hunting the correct code now.

jnm2:
This seems to be close to what I need. Usage:

```cs
[Test]
public static void Default_configuration_is_valid()
{
    // See https://github.com/castleproject/Windsor/issues/328

    using (var container = Program.InitializeContainer())
    {
        if (TryGetUnresolvableDependenciesMessage(container.Kernel, out var message))
            Assert.Fail(message);
    }
}
```

Implementation:

```cs
private static bool TryGetUnresolvableDependenciesMessage(IKernel kernel, out string message)
{
    var actuallyUnresolvable = new List<(DependencyModel dependency, IHandler handler)>();

[...]

PR #376, opened 2018-01-27
ASP.NET Core facility

fir3pho3nixx:
This addresses https://github.com/castleproject/Windsor/issues/120

It provides an integration with aspnet core 2.0. 

Considerations for this PR:

 - Windsor is a non-conforming container. This facility implements the "2 container" solution. 

The vanilla aspnet container consuming "Framework Configured Container" via the  IServiceCollection and IApplicationBuilder interfaces.

The windsor integration as the "Application Container". These are well known framework types that serve as integration points for application developers provided by ASPNET Core 2.0. 

The "application container" integrates with the "framework configured container" by using a sub dependency resolver([FrameworkConfigurationDependencyResolver](https://github.com/castleproject/Windsor/pull/376/commits/d731120d78f6d8ae393b9d2c08a04a1adf12422e#diff-7f7ad5558608eecf569559d9cfc66a9c)). This allows seamless type resolution of framework configured components without having to "[cross wire](https://github.com/simpleinjector/SimpleInjector/issues/437)" things. 

"Cross wiring" at the time of writing was attempted by SimpleInjector but it led to side effects ie. "Captive Dependencies" and "Torn Lifestyles". 

  - This is also a preview of the NUnit Adapter Upgrade. 

We have an issue here: https://github.com/castleproject/Windsor/issues/243, which this hopefully addresses in how we run tests, we can switch over to 'dotnet test'.  Please see [this](https://github.com/castleproject/Windsor/pull/376/commits/d731120d78f6d8ae393b9d2c08a04a1adf12422e#diff-2a6b8433c8e101b29a4fc81759dcf189R7) and [this](https://github.com/castleproject/Windsor/pull/376/commits/bc2d2cb496a09a5c8c5f1a636bb44642a6a2c117#diff-c829783b483aac27eed49629a0fecbb6R72)

Would be good if we could roll this out across the board. It will also start exposing the various target frameworks(net45, netcoreapp1.0, netcoreapp2.0) in VS2017/Resharper(latest) which is handy. 

 - This would not have been possible without all the work done by @hikalkan and @dotnetjunkie. 

These guys both tried to implement the Microsoft Extensions Dependency Injection Abstraction(myself included but gave up). Credit for creating a foundation from which to learn from! :)


appveyor.yml
```diff
@@ -17,24 +17,7 @@ init:
...
-    $wc.UploadFile("https://ci.appveyor.com/api/testresults/nunit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path "src\Castle.Windsor.Tests\$testBinPath\TestResult_Windsor.xml"))
-
-    $wc.UploadFile("https://ci.appveyor.com/api/testresults/nunit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path "src\Castle.Windsor.Tests\$testBinPath\TestResult_Windsor_NetCore.xml"))
-
-    $wc.UploadFile("https://ci.appveyor.com/api/testresults/nunit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path "src\Castle.Facilities.WcfIntegration.Tests\$testBinPath\TestResult_WcfIntegration.xml"))
-
-    Push-AppveyorArtifact $env:APPVEYOR_BUILD_FOLDER\src\Castle.Windsor.Tests\$testBinPath\TestResult_Windsor.xml
-
-    Push-AppveyorArtifact $env:APPVEYOR_BUILD_FOLDER\src\Castle.Facilities.WcfIntegration.Tests\$testBinPath\TestResult_WcfIntegration.xml
-
```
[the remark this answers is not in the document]
fir3pho3nixx: It was also causing pathing issues that could not be picked up locally. Bad reason for a build to fail :'(

src/Castle.Windsor/MicroKernel/IKernel.cs
```diff
@@ -148,6 +148,13 @@ IKernel AddFacility<T>(Action<T> onCreate)
 		/// <returns></returns>
 		IHandler GetHandler(String name);
 
+		/// <summary>
+		///   Returns the <see cref = "IHandler" />
+		///   for the all components.
+		/// </summary>
+		/// <returns>Handlers which implement sub dependency resolvers</returns>
+		IHandler[] GetHandlers();
```
fir3pho3nixx: @jnm2 - I have made getting all handlers easy. This is going to start playing into validation of configurations if @jonorossi approves this PR.

fir3pho3nixx: I needed to validate against "framework configuration components" accidentally registered in the container here: https://github.com/castleproject/Windsor/pull/376/commits/463da9f141e8a0fa8c87d066679ae5e028995796

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -48,5 +51,18 @@ public static void UseCastleWindsor<TStartup>(this IApplicationBuilder app, IWin
...
+		{
+			ThrowIfFrameworkComponentsAreRegistered(container);
+		}
+
+		private static void ThrowIfFrameworkComponentsAreRegistered(IWindsorContainer container)
+		{
+			var handlers = container.Kernel.GetHandlers();
+			var hasAspNetCoreFrameworkRegistrations = handlers.Any(x => x.ComponentModel.Implementation.FullName.StartsWith("Microsoft.AspNetCore"));
+			if (hasAspNetCoreFrameworkRegistrations)
+				throw new Exception("Looks like you have implementations registered from 'Microsoft.AspNetCore'. Please do not do this as it could lead to torn lifestyles. Please remove the registrations.");
```
fir3pho3nixx: I should probably rename the exception message to:

Looks like you have implementations registered from 'Microsoft.AspNetCore'. Please do not do this as it could lead to torn lifestyles. Please remove the registrations from Castle Windsor.

src/Castle.Windsor.Tests/Pools/PooledLifestyleManagerTestCase.cs
```diff
@@ -462,6 +462,11 @@ public IHandler GetHandler(string name)
 				throw new NotImplementedException();
 			}
 
+			public IHandler[] GetHandlers()
+			{
+				throw new NotImplementedException();
+			}
+
```
fir3pho3nixx: Just followed the pattern that was there already. Additive change.

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -426,6 +426,31 @@ public virtual IHandler[] GetHandlers(Type service)
...
+				{
+					var newResult = new IHandler[result.Length + parentResult.Length];
+					result.CopyTo(newResult, 0);
+					parentResult.CopyTo(newResult, result.Length);
+					result = newResult;
+				}
+			}
+
+			return result;
+		}
```
fir3pho3nixx: @jnm2 - I have made [getting all handlers](https://github.com/castleproject/Windsor/pull/376/commits/463da9f141e8a0fa8c87d066679ae5e028995796#diff-ff15a9919978ddd9419b7185d67a319dR433) easy. This is going to start playing into [validation of configurations for tests](https://github.com/castleproject/Windsor/issues/328) if @jonorossi approves this PR.

I needed to validate against "framework configuration components" accidentally registered in the container here: https://github.com/castleproject/Windsor/pull/376/commits/463da9f141e8a0fa8c87d066679ae5e028995796

fir3pho3nixx:
@jonorossi - If you could take a look that would be great.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+			container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly).BasedOn<TagHelper>().LifestyleTransient());
+		}
+
+		public static void UseCastleWindsorMiddleware<T>(this IApplicationBuilder app, IWindsorContainer container) where T : class, ICastleWindsorMiddleware
+		{
+			container.Register(Component.For<T>());
+			app.Use(async (context, next) => { await container.Resolve<T>().Invoke(context, next); });
+		}
+
+		public static void ValidateConfiguration(this IWindsorContainer container)
```
jonorossi: This extension method sounds sort of misleading extending `IWindsorContainer` as it gets called looking like `container.ValidateConfiguration();`. Also do users have to call this manually as your example in the docs doesn't show it and only a unit test calls it.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+		{
+			container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly).BasedOn<Controller>().LifestyleScoped());
+			container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly).BasedOn<ViewComponent>().LifestyleTransient());
+			container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly).BasedOn<TagHelper>().LifestyleTransient());
+		}
+
+		public static void UseCastleWindsorMiddleware<T>(this IApplicationBuilder app, IWindsorContainer container) where T : class, ICastleWindsorMiddleware
+		{
+			container.Register(Component.For<T>());
+			app.Use(async (context, next) => { await container.Resolve<T>().Invoke(context, next); });
```
jonorossi: Does calling `Invoke` directly here rather than returning the middleware cause any problems with the framework? I'm an ASP.NET Core newbie but I just wondered if there was maybe a diagnostic trace that it keeps state of as each middleware is executed.

src/Castle.Facilities.AspNetCore/ICastleWindsorMiddleware.cs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+namespace Castle.Facilities.AspNetCore
+{
+	using System;
+	using System.Threading.Tasks;
+
+	using Microsoft.AspNetCore.Http;
+
+	public interface ICastleWindsorMiddleware
```
jonorossi: Could you explain why this couldn't use `IMiddleware`? Does ASP.NET Core automatically wire up `IMiddleware`s not allowing them to be resolved from the container?

It looks like ASP.NET Core already handles this, see the ["Per-request dependencies" section](
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware?tabs=aspnetcore2x), is this functionality useful?

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+			services.AddCustomTagHelperActivation(container.Resolve);
+			services.AddCustomViewComponentActivation(container.Resolve);
+			container.Kernel.Resolver.AddSubResolver(new FrameworkConfigurationDependencyResolver(services));
+		}
+
+		public static void UseCastleWindsor<TStartup>(this IApplicationBuilder app, IWindsorContainer container)
+		{
+			container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly).BasedOn<Controller>().LifestyleScoped());
+			container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly).BasedOn<ViewComponent>().LifestyleTransient());
+			container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly).BasedOn<TagHelper>().LifestyleTransient());
```
jonorossi: If your controllers are in a different assembly do you want people register their components themselves and skip `UseCastleWindsor`? You mentioned in the docs if they wanted different lifestyles but not if they are in different assemblies.

XML comments on these with what they specifically do would help people saving having to jump to the docs or into the code.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+	{
+		private static readonly string[] InvalidFrameworkConfigurationNamespaces =
+		{
+			"Microsoft.AspNetCore"
+			// Todo: Might need to add some more framework namespaces.
+		};
+
+		public static void AddCastleWindsor(this IServiceCollection services, IWindsorContainer container)
+		{
+			services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```
jonorossi: I don't see `IHttpContextAccessor` used in the facility, just wanting to understand what it is needed for.

appveyor.yml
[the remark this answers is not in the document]
jonorossi: Agreed that NUnit's console output is much better for failures now so we don't need the last 2, however the "/api/testresults/nunit/..." ones are for the "Tests" tab in AppVeyor. Castle Core only has the first ones.

generik0:
HI. 
Great faciltiy. Just what i was needing.

Could it be interesting to add the Microsoft.Extensions.Logging as a resolver (or something) into the faciltiy? I know this would mean a dependancy on the Microsoft.Extensions.Logging but it is inbuilt and automatically registerd with aspnetcore.

```
public class LoggerResolver : ISubDependencyResolver
    {
        private readonly ILoggerFactory _loggingFactory;

        public LoggerResolver(ILoggerFactory loggingFactory)
        {
            _loggingFactory = loggingFactory;
        }

        public bool CanResolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model,
            DependencyModel dependency)
        {
            return dependency.TargetType == typeof(ILogger);
        }

        public object Resolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model,
            DependencyModel dependency)
        {
            return _loggingFactory.CreateLogger(model.Name);
        }
    }
```c#

**Comment by fir3pho3nixx** [comment]
@generik0 

We are using the following meta NuGet package:

```xml
<PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.5" />
```

Had a quick look and this did include: 

```xml
<PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.0" />
```

In my original repo I also managed to inject LoggingFactory into a controller as a test. 

```csharp
public HomeController(IUserService userService, ILoggerFactory loggerFactory, IViewBufferScope viewBufferScope)
{
	this.userService = userService;
	this.loggerFactory = loggerFactory;
	this.viewBufferScope = viewBufferScope;
}
```

Code is over here: https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/master/WebApp/Controllers/HomeController.cs#L15

You should be able to easily inject your framework types without having to register them in Windsor or using the sub dependency resolver. If this is not happening or not working I would love to see an example so I can fix it.

appveyor.yml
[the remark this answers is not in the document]
fir3pho3nixx: Does anyone look at those test tabs? I dont. Also noticed test output from netcoreapp1.0 would wipe out test results from net45 because of the namespace collisions. Let me know if you use this and I will put it back.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Breaking changes:
 - Created Castle.Facilities.AspNet.SystemWeb facility so we can remove this from the Windsor core library. (@fir3pho3nixx, #283)
 
 New Features:
+- Created Castle.Facilities.AspNetCore facility to support Mvc web applications using dotnet core. (@fir3pho3nixx, #120)
```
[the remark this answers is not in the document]
fir3pho3nixx: We should try, let me double check this and get back to you.

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>netcoreapp2.0</TargetFramework>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.5" />
```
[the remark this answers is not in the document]
fir3pho3nixx: I can cut this right back. 

I just thought it would be handy if you created a console app, then installed Castle.Factilities.AspNetCore you would have everything ready go as a reference.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: This is a helpful exception, that could hopefully be added to a test fixture somewhere. Did not want this getting in the way of having to register framework types(even though we say you should not) but in a test scenario it should light up red. The docs needs a bit updating here for sure. Let me write this up a bit better.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		public static void ValidateConfiguration(this IWindsorContainer container)
+		{
+			ThrowIfFrameworkComponentsAreRegistered(container);
+		}
+
+		private static void ThrowIfFrameworkComponentsAreRegistered(IWindsorContainer container)
+		{
+			var offendingFrameworkTypes = InvalidFrameworkConfigurationNamespaces;
```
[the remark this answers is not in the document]
fir3pho3nixx: > The variable name refers to type names which you prime with namespace names than compare against Type.FullName, obviously FullName returns the namespace but if this should only match the namespace then it should probably look at namespace otherwise should be renamed.

Will do the namespace option. 

 > I wouldn't have a problem with this being hardcoded compared against a single string rather than "building for the future".

Will move it back to how I had it before. :D

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+			foreach (var invalidFrameworkConfiguredType in offendingFrameworkTypes)
+			{
+				var handlers = container.Kernel.GetHandlers();
```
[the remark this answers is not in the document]
fir3pho3nixx: Yup, that is garbage code. Will fix.

generik0:
@fir3pho3nixx 
Yes i know i can inject the factory. but even better to inject the ILogger directly. Hence, there is even a better reason to add the resolver in your spnetcore Faciltiy
Allowing for:
``c#
public HomeController(IUserService userService, ILogger logger, IViewBufferScope viewBufferScope)
{
	this.userService = userService;
	this.logger= logger;
	this.viewBufferScope = viewBufferScope;

	this.logger.LogDebug("Log something baby");
}
```

**Comment by fir3pho3nixx** [comment]
@generik0 - Let me check this and get back to you. It should support ILogger.

**Comment by generik0** [comment]
@fir3pho3nixx

If i write this.loggibgfactory.CreateLogger(this.GetType()); 
Then i get a logger, you need to pass the calking type to the Factory.

Hence i made the resolver to ADD info you facilitet :-)

If you want to include this feature i Can easily ADD as a pull request :-)

**Comment by fir3pho3nixx** [comment]
@generik0

Can you try again with the code here https://github.com/fir3pho3nixx/Windsor.AspNetCore. I have enabled the delegation of open/closed generics. This now allows constructors like this to be created: 

```csharp
public HomeController(IUserService userService, ILoggerFactory loggerFactory, ILogger<HomeController> logger, IViewBufferScope viewBufferScope, IOpenGenericService<ClosedGenericTypeParameter> closedGenericService)
{
	this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
	this.userService = userService ?? throw new ArgumentNullException(nameof(userService));
	this.loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
	this.viewBufferScope = viewBufferScope ?? throw new ArgumentNullException(nameof(viewBufferScope));
	this.closedGenericService = closedGenericService;
}
```

Following behaviours are supported:

 - `ILogger'T` is an open generic service. Resolves.
 - `IOpenGenericService<ClosedGenericTypeParameter>` is closed generic service. Resolves.

Most of the change was [here](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/797a697e3d88695af934470b03ba5c6211feb5fb#diff-2a2dbcb40447e78d1b3cc007d47740f7)

If there are any more glaring obvious things you can spot, by all means PR away to my forked remote. Thanks for helping us find this. Will commit this change with tests tomorrow night :)

generik0:
@fir3phonixx
That’s great. Always nice with a generic resolver.
I have to admin though. I am to lazy to add a generic to a constructor injected object with the same type that the main class is. I.e. you wrote HomeController twice, the name of the class and the in the ILogger. This will really irritate me in the long run.

the resolver should do all the work for me.
Hence, I personally believe the loggerresolver I wrote above would be better to add.
No stress though, I can always add it myself in my own code :-)

fir3pho3nixx:
Yup, I could see how that could be irritating.  

Please submit PR here: https://github.com/fir3pho3nixx/Windsor/tree/facilities-aspnetcore

:)

jonorossi:
> I have to admin though. I am to lazy to add a generic to a constructor injected object with the same type that the main class is. I.e. you wrote HomeController twice, the name of the class and the in the ILogger. This will really irritate me in the long run.

I agree, `ILogger<HomeController> logger` just looks strange and prone to error, glad you guys are looking at changing that.

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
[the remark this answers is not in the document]
jonorossi: Yes, I think we should cut it back.

The ASP.NET team just [announced today that they are actually introducing a new `Microsoft.AspNetCore.App` package](https://github.com/aspnet/Announcements/issues/287) and are changing their new project templates not to use ".All" in favour of this one because the "All" metapackage actually brings in heaps of stuff that isn't expected like Azure KeyVault, a Redis client and Sqlite. We really shouldn't force others to bring in packages we don't need for convenience.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
jonorossi: Thanks for the description, I can see how this would be helpful for a unit test, I just expected that it would also get throw during runtime as most users especially those that are new aren't likely to be unit testing this stuff. But I could see there might be a reason someone really does need to register one of these types, and it is probably a good idea not to block it hard.

appveyor.yml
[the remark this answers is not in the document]
jonorossi: Good point, the test tab is pretty pointless because I don't think you can ever navigate to anything including failures unlikely TeamCity.

It doesn't look like one test run wipes out the other, there are duplicate test names under different file names, `Castle.Windsor.Tests.dll` (.NET 4.5) and `Castle.Windsor.Tests.exe` (.NET Core).

But yer I agree let's just get rid of them. Should we also run NUnit's runner with `--noresult` since the command line output is plenty good?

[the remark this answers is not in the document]
fir3pho3nixx: https://github.com/castleproject/Windsor/pull/376/commits/336fdc8a2eab972aaccbd18fc11d11c88302b106

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: > If your controllers are in a different assembly do you want people register their components themselves and skip UseCastleWindsor? You mentioned in the docs if they wanted different lifestyles but not if they are in different assemblies.

I updated the docs to be a bit clearer. 

> XML comments on these with what they specifically do would help people saving having to jump to the docs or into the code.

Also added some XML comments.

src/Castle.Facilities.AspNetCore/ICastleWindsorMiddleware.cs
fir3pho3nixx: > Could you explain why this couldn't use IMiddleware? Does ASP.NET Core automatically wire up IMiddlewares not allowing them to be resolved from the container?

Great question. You can definitely do this but you are moving away from the Windsor container and falling back onto "Framework Configured Container". This creates problems if your middleware consumes dependencies that are know to Windsor but not registered within the IServiceCollection. If anyone decided to go down this route they will end up with a mish mash of registrations across both containers. This should be avoided at all costs. 

> It looks like ASP.NET Core already handles this, see the "Per-request dependencies" section, is this functionality useful?

If their middleware does not consume any dependencies, then this functionality is useful. Otherwise it is recommended you use our extension because then you are free to consume your own types(registered in Windsor) and framework types which is dealt with in the sub resolvers.

I did update the XML documentation for this extension.

fir3pho3nixx:
@jonorossi - This is now ready for one final check.

fir3pho3nixx:
@jonorossi - This is now ready for one final check.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: > Does calling Invoke directly here rather than returning the middleware cause any problems with the framework? 

Not that I have come across. Do you have any further links?

fir3pho3nixx:
@jonorossi - This is now ready for one final check.

fir3pho3nixx:
@jonorossi - This is now ready for one final check.

fir3pho3nixx:
@jonorossi - This is now ready for one final check.

fir3pho3nixx:
@jonorossi - This is now ready for one final check.

fir3pho3nixx:
@jonorossi - This is now ready for one final check.

fir3pho3nixx:
@jonorossi - I also double checked there were no undue memory leaks. I even introduced [console and debug logging](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/ad94ba48e80e8b2f72fbb3e705a7f99887483762). I did find the fairness from the CPU was a bit weird with netcore apps.

I also introduced [IDisposable on the AspNetUserService](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/dde09a54c68f710da62785755752010587b4590a#diff-68d907e65062d369b9f82e62dd896c6bR69) just to be sure burdens and scopes are behaving themselves.

fir3pho3nixx:
Should check this on ubuntu linux.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
jonorossi: Nope, it was just a thought as I've never dug into how this stuff works.

jonorossi:
> Should check this on ubuntu linux before merge.

I guess that should be another pull request since Windsor doesn't have a Travis CI build like Castle Core.

docs/facilities.md
```diff
@@ -43,6 +43,7 @@ In addition to the above, as part of Castle Project, some other facilities are p
 * [System Web Facility](systemweb-facility.md) - Provides system web integration for web projects using `PerWebRequest` lifestyles.
 * [AspNet Mvc Facility](aspnetmvc-facility.md) - Provides aspnet mvc integration for web projects using Windsor.
 * [AspNet WebApi Facility](aspnetwebapi-facility.md) - Provides aspnet webapi integration for web projects using Windsor.
+* [AspNet Core Facility](aspnetcore-facility.md) - Provides aspnet core integration for web projects using Windsor.
```
jonorossi: "ASP.NET"

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
```diff
@@ -0,0 +1,31 @@
...
+	<PropertyGroup>
+		<TargetFramework>netstandard2.0</TargetFramework>
+	</PropertyGroup>
+
+	<Import Project="..\..\buildscripts\common.props"></Import>
+
+	<PropertyGroup>
+		<GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+		<PackageId>Castle.Facilities.AspNetCore</PackageId>
+		<Title>Castle Windsor AspNet Core facility</Title>
```
jonorossi: "ASP.NET Core"

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
```diff
@@ -0,0 +1,31 @@
...
+		<PackageTags>castle, windsor, inversionOfControl, DependencyInjection, aspnet, core</PackageTags>
+		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+		<AssemblyName>Castle.Facilities.AspNetCore</AssemblyName>
+		<RootNamespace>Castle.Facilities.AspNetCore</RootNamespace>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Microsoft.AspNetCore" Version="2.0.1" />
+		<PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.0.2" />
+		<PackageReference Include="Microsoft.AspNetCore.Mvc.Razor" Version="2.0.2" />
```
jonorossi: I'm surprised this facility needs ASP.NET Core MVC and Razor, and not just the main ASP.NET Core package. I haven't had a chance to check out the work the Nancy guys have done to get Nancy running on ASP.NET Core but I thought this facility might be a drop in.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Breaking changes:
 - Created Castle.Facilities.AspNet.SystemWeb facility so we can remove this from the Windsor core library. (@fir3pho3nixx, #283)
 
 New Features:
+- Created Castle.Facilities.AspNetCore facility to support Mvc web applications on dotnet core. (@fir3pho3nixx, #120)
```
jonorossi: Did you get a chance to try this out using ASP.NET Core on .NET Framework?

src/Castle.Facilities.AspNetCore/ICastleWindsorMiddleware.cs
jonorossi: This still feels icky as it means your application code has to implement a container specific interface otherwise the container won't let you use it. It feels a bit like the lifestyle attributes Windsor has that aren't really recommended as registration/configuration should happen outside the application code, or the anti-pattern that is depending on `IKernel` from non-factory components.

If a user did implement `IMiddleware` or used someone else's middleware would ASP.NET Core automatically wire it up, or can we just tell them not to register it with ASP.NET Core and instead register it with the Windsor facility?

I think that the type name (`Castle.Facilities.AspNetCore.ICastleWindsorMiddleware`) is something that keeps sticking its nose at me. Having the product name in the interface name is something I don't believe we've ever had to do before which sort of tells me we are doing something wrong, especially that Windsor doesn't have middleware but that this is a workaround.

> I did update the XML documentation for this extension.

FYI, the interface doesn't have any documentation?

CHANGELOG.md
jonorossi: If not, feel free to just log an issue so we don't forget for v5 and defer it for now.

jonorossi:
@fir3pho3nixx sorry for sitting on your comments the last few days, I'll try to be more responsive. A few minor issues but that middleware interface is still a sticking point for me. We are nearly there 🤞.

src/Castle.Facilities.AspNetCore/ICastleWindsorMiddleware.cs
fir3pho3nixx: > This still feels icky as it means your application code has to implement a container specific interface otherwise the container won't let you use it.

You are right it is superfluous and can be removed altogether. 

 > If a user did implement IMiddleware or used someone else's middleware would ASP.NET Core automatically wire it up, or can we just tell them not to register it with ASP.NET Core and instead register it with the Windsor facility?

Not aware of anything that wires this up for you. 

For framework middleware:

 - [You still have to "tell" the IServiceCollection the middleware exists as a resolvable type](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/15f5833a4b097f825f223cc788e3564bbdc88dd4#diff-68d907e65062d369b9f82e62dd896c6bR38).

 - [You have to register the same type as middleware with the app builder](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/15f5833a4b097f825f223cc788e3564bbdc88dd4#diff-68d907e65062d369b9f82e62dd896c6bR54).  

 - It does not resolve types registered in Windsor, if you need this then register it in your container and use [this extension method instead](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/15f5833a4b097f825f223cc788e3564bbdc88dd4#diff-573837522f7d6c1d750af3ef541e4629R70).

 > I think that the type name (Castle.Facilities.AspNetCore.ICastleWindsorMiddleware) is something that keeps sticking its nose at me. 

Let's delete it. We still need the extension method though right?

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
fir3pho3nixx: Razor is there because of the activator for [TagHelpers](https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/master/Castle.Facilities.AspNetCore/Activators/DelegatingTagHelperActivator.cs#L19).

Mvc is there because of the activator for [Controllers](https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/master/Castle.Facilities.AspNetCore/Activators/DelegatingControllerActivator.cs#L19).

Open to any idea's that could make this easier.

CHANGELOG.md
fir3pho3nixx: I did check it for all of about 2-3 minutes before moving on to something else. Let me get back to you on this one. Let's hold out before raising the issue.

fir3pho3nixx:
> @fir3pho3nixx sorry for sitting on your comments the last few days, I'll try to be more responsive. 

Don't worry. 

 > A few minor issues but that middleware interface is still a sticking point for me. 

Glad it is. I can delete the interface. Let me get those changes in tomorrow night and then we can evaluate what else we need then.

 > We are nearly there 🤞.

Time to squish me thinks.

src/Castle.Facilities.AspNetCore/ICastleWindsorMiddleware.cs
jonorossi: Yay, `ICastleWindsorMiddleware` is gone 🎉 .

Yes, we'll still need the extension method, you were saying trying to do it via the framework container will just cause too much pain so happy for it to stay, however I noticed you renamed the `IApplicationBuilder` extension method:

```
app.UseCastleWindsorMiddleware<CustomMiddleware>(container);
... to ...
app.UseResolvableMiddleware<CustomMiddleware>(container);
```

I wonder if this new name is very clear, if you didn't want "Windsor" in the extension method name (which I thought was convention for these IApplicationBuilder thingys) then naming it just `UseMiddleware` like the framework one would explain just as much but you'd have an overload with the container.

```
app.UseMiddleware<CustomMiddleware>(); // register with framework
app.UseMiddleware<CustomMiddleware>(container); // register with Windsor
```

Or maybe name it `app.UseMiddlewareViaWindsor<CustomMiddleware>(container)`?

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
jonorossi: > Open to any idea's that could make this easier.

I've got no ideas, I think just leave this as is for now. Nancy is different with its container usage anyway wired through the bootstrapper so let's get this going and the first person to want Windsor on Nancy can sort that one out.

jonorossi:
> Time to squish me thinks.

Yep. Let me know how .NET Framework goes.

generik0:
Hi

I am using it on .Net Framework. I.e. I have copied the files over.
Project just starting so there is not much in it, but the framework seams to be integrating nicely...

CHANGELOG.md
fir3pho3nixx: @jonorossi - I looked at this again, it is definitely possible to run ASP.NET Core targeting [net461](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/c21d456022294f5a128b5606872eced2f39f8485#diff-bc4f9cd406d7279d4b85cfeaff4a5403R1).  

The Visual Studio WebApp template appears to have a NuGet reference to [Microsoft.AspNetCore.All](https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/c21d456022294f5a128b5606872eced2f39f8485#diff-bc4f9cd406d7279d4b85cfeaff4a5403L8) which misleadingly targets netcoreapp2.0 instead of netstandard2.0. Once I added in the individual NuGets I needed it ran without any issues.

fir3pho3nixx:
> Yep. Let me know how .NET Framework goes.

Please see comments I left [here](https://github.com/castleproject/Windsor/pull/376#discussion_r166738236)

fir3pho3nixx:
@jonorossi - Think we are good to go.

CHANGELOG.md
jonorossi: > The Visual Studio WebApp template appears to have a NuGet reference to Microsoft.AspNetCore.All which misleadingly targets netcoreapp2.0 instead of netstandard2.0. Once I added in the individual NuGets I needed it ran without any issues.

The `Microsoft.AspNetCore.All` package [by design](https://github.com/aspnet/Announcements/issues/243) is .NET Core only, I think also one of the reasons they have introduced `Microsoft.AspNetCore.App`.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+				await resolve.InvokeAsync(context, async (ctx) => await next());
+			});
+		}
+
+		/// <summary>
+		/// Optional helpful exception to validate against types being installed into Windsor from the 'Microsoft.AspNetCore'
+		/// namespace. This guards against 'Torn Lifestyles' and 'Captive Depenencies'.
+		/// </summary>
+		/// <param name="container">Windsor container</param>
+		public static void ValidateConfiguration(this IWindsorContainer container)
```
jonorossi: @fir3pho3nixx I just realised this is an extension method on `IWindsorContainer`, I think this will confuse users when they import the `Castle.Facilities.AspNetCore` namespace as the usage will appear:

```
container.ValidateConfiguration();
```

And it doesn't validate configuration for the whole container, just for this facility.

I'm going to merge this pull request, but I think it would be a worthwhile fix to scope this either by changing the method name or something else.

fir3pho3nixx: I will slot this into a separate PR. How about renaming it to ValidateNoFrameworkComponentsAreRegistered?

generik0:
Hi
Is it supposed to be that i:
c#```
internal void ConfigureServices(IServiceCollection services)
        {
            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
            services.AddMvcWithJson();
            services.AddCastleWindsor(_container);
        }
```

i.e register the accessor?
` services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();`
Shouldn't this be done in the factory?

**Comment by generik0** [comment]
Hi again
Why do i have to:
`app.UseCastleWindsor<Program>(_container);`
why not just
`app.UseCastleWindsor(_container);`
And it uses the calling assembily?

**Comment by fir3pho3nixx** [comment]
@generik0

 > IHttpContextAccessor
 
What problems are you having with this?

Does this help? 

https://stackoverflow.com/questions/38184583/how-to-add-ihttpcontextaccessor-in-the-startup-class-in-the-di-in-asp-net-core-1

 > Why do I have to register things using: `app.UseCastleWindsor<Program>(_container);`? Why not just use `app.UseCastleWindsor(_container);` and then rely on `GetEntryAssembly` for conventional component registrations?

Good question. I think you have spotted a hole in our documentation. You are right, I could have done it this way. The reason I chose not to is for scalability. 

Imagine this scenario:

Your app could be so separated out that you need different assemblies for `view components`, `tag helpers` or `controllers`. 
It was with this in mind I decided passing in the type as a generic for `app.UseCastleWindsor<Program>(_container);` would create a way of overloading this construct for independent assemblies. 

The problem with `GetEntryAssembly` is that it assumes all these types are defined in that assembly. This could become a problem if your code is declared in another assembly.

Here is an example of what I mean:

app.UseCastleWindsor<ViewComponent1>(_container); // <- Defined in assembly My.Acme.ViewComponents.csproj
app.UseCastleWindsor<TagHelper1>(_container);  // <- Defined in assembly My.Acme.TagHelpers.csproj
app.UseCastleWindsor<Controller1>(_container);  // <- Defined in assembly My.Acme.Controllers.csproj

Look forward to hearing back.

**Comment by generik0** [comment]
Hi @fir3pho3nixx

I am a strong believer of the api and fluent registration. There is the problem when you have interface assemblies for everything, then the concrete assemblies are not refernced. But this can be overcome simply.

Regarding the app.UseCastleWindsor(_container) why not allow both methods? Running normally and running with custom assemblies.

I can make a pull request with some tests :-)

**Comment by generik0** [comment]
Hi @fir3pho3nixx


I don’t understand why the castle aspnetcore facility is not registering the IHttpContextAccessor. I have e.g. used Aufofac and it registers it automatically.

Maybe there should be a new call to the facility to register this feature?

**Comment by fir3pho3nixx** [comment]
> Regarding the app.UseCastleWindsor(_container) why not allow both methods? Running normally and running with custom assemblies.

I am definitely open to suggestions here, can you go ahead and raise a PR?

 > I don’t understand why the castle aspnetcore facility is not registering the IHttpContextAccessor. I have e.g. used Aufofac and it registers it automatically. Maybe there should be a new call to the facility to register this feature?

I have just started using this facility in a project at work. Let me see if I can get it registered there and I will let you know if I have any problems.

**Comment by generik0** [comment]
@fir3pho3nixx
Thats great. BRB with PR

**Comment by fir3pho3nixx** [comment]
@generik0 

I had a look at injecting IHttpContextAccessor as a service and could not find any problems. 

[This](https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/context-accessor/WebApp/Startup.cs#L40) is where I registered it and [this](https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/context-accessor/WebApp/Controllers/HomeController.cs#L22) is where I resolved it. All runs fine. 

Do you have a repro?

**Comment by generik0** [comment]
Hi

Are we should there should be dependancies on AspNetCore.Mvc.Razor ? 
A simple WebApi project does not use razor.

From WebApi template:
image

**Comment by fir3pho3nixx** [comment]
@generik0 - The razor dependency is there for view components and tag helpers. Don’t think we can remove it if we want to support MVC.

**Comment by generik0** [comment]
@fir3pho3nixx I know why it is there. I am just thinking if we could do it differently.
even though it wont help the dependancies, Autofac e.g. has sepater helpers to register controlleres, accessor etc:
Should we always register the controllers, helpers and everything, or is it better to split?
Maybe both possibilites could be provided. I could extend to ahve individual registrations also?
image

**Comment by fir3pho3nixx** [comment]
@generik0 - the castle windsor community has been waiting for so long for an aspnet core facility, I have personally been researching this problem for over a year now. Before we prematurely optimise this by applying separation of concerns perhaps we should make the entry easy breezy before we start breaking this out into different NuGet packages. 

I am not saying you are wrong, you are right. What I am saying is, let's keep it slim to start(which might mean breaking some SoC rules) and let the community decide as we go. What you are suggesting creates more documentation, more NuGet's and honestly this is not something I want for people out there as a first go. The templates for ASP.NET Core that come with VS2017 are rubbish, but that is where people start and unfortunately it includes references to meta NuGet packages like [Microsoft.AspNet.Core.All](https://www.nuget.org/packages/Microsoft.AspNetCore.all). 

The barrier to entry should be kept low. We can evolve this later on. Not even Microsoft has solved this problem fully yet. Look forward to your response.

**Comment by generik0** [comment]
Hi. Should @inject be working on the views?
I have one special place needed :-/

**Comment by fir3pho3nixx** [comment]
Yes it should! I will check this out before we release. Good catch!

**Comment by generik0** [comment]
It is one of web college that has used it.
It was failing for him. I will look at it tonight, it very might not be castle, but rather him...
I will get back to you ASAP!

**Comment by generik0** [comment]
Hi @fir3pho3nixx 
There does appear to be an issue with a custom regisered service that can be injected into a controller. It is not working in the view with @inject. However a MS standard is e.g. IHtmlLocalizer :


image


image

image

**Comment by generik0** [comment]
Hi @fir3pho3nixx 
I have spent a couple of hours on it. Even is i register the custom service and class for the service, i.e. service.AddSingleton<x, ix> it doesn't work...

Any ideas. I am at a loss...

**Comment by generik0** [comment]
Can't do a workaround and inject into the model property either
image

**Comment by fir3pho3nixx** [comment]
@generik0 - I will have a look. Give me 1/2 days I will get back to you.

**Comment by generik0** [comment]
@fir3pho3nixx Any luck. I have been trying to figure it out also. It gives no meaning when e.g. IHtmlLocalizer can be @inject

**Comment by fir3pho3nixx** [comment]
I need a bit more time on this one, there is no tidy way of achieving this. There no activators and classes in this area are all sealed. When I get back to London I will see if I can implement a custom razor feature that does this.

**Comment by generik0** [comment]
ok, No problem
I dont think the razorpage is such hi priority. I have not tested with Autofac either. This was the old asp.net way to inject into a view (i.e. though the model. But the direct injection of a interface. i.e. @Inject ICqrsDispatcher Dispatcher on the view, is very important.

Maybe we could do a xp session together and look into it if you like?

**Comment by fir3pho3nixx** [comment]
Will have to wait until I get back to London. Currently in Japan. Will email you when I am back to set something up 👍

**Comment by fir3pho3nixx** [comment]
@generik0 

I have investigated this extensively over the weekend and came to the conclusion that the only way we can achieve this is by implementing a new "Razor Feature/Directive". 

I followed the path of how the existing inject directive works: 
https://github.com/aspnet/Razor/blob/2f79b90af5cbd9499f012c204f5bf4bb413ae61b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensions.cs#L20

Any extensibility like this is offered through the RazorViewEngineOptions:
https://github.com/aspnet/Mvc/blob/dfa085afaf4ddb0b1063eed35e070d6189acaca2/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs

Unfortunately there is no interface for adding new Razor directives, there is also simply no public API for getting hold of the IRazorEngineBuilder either(which allows the adding of these features) when the IRazorEngine get's new'ed up. Specifically something like this:
https://github.com/aspnet/Razor/blob/0c6ec099584798b5222b669633604269c75d5f15/src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs#L27

This kind of leaves in a place where we have to create a custom RazorViewEngine that emulates the existing features/directives of Microsoft plus the additional one for doing Castle Windsor compliant @inject's. 

After raising this issue here: 
https://github.com/aspnet/Mvc/issues/7540

I also came to the conclusion that the tooling around this would lag behind and also produce errors when it really shouldn't. 

To make the @inject work with the project in it's current form, you would have to cross wire your dependencies into the ServiceCollection. This is yukky and I would never recommend people go down this path. 

Here is also an issue I found with how SimpleInjector have approached this:
https://github.com/simpleinjector/SimpleInjector/issues/362

I am interested to know how static service location would solve the problem, have you tried this?

**Comment by generik0** [comment]
Hi. Great work!. Sorry it ended a little in a dead end. I guess We know now why autofac retur an IServiceCollector og the service init og startup. They re wrote the IServiceCollector, which i agere is not a viable solution.

Thanks for the simple injector link. Are you saying That this Will work:

services.AddScoped<ISomething>(provider => container.GetInstance<ISomething>());

Right now using the view bag, so anything would be better!!!

**Comment by generik0** [comment]
This really is bad design by ms aspnetcore

**Comment by generik0** [comment]
HI @fir3pho3nixx 
This seams to work. Its fine / great.! No problem at all. I tried with an installer, but this did not work. But i ended up with this:

Questions:

- Would  it be better doing this as middleware?
- As i see it one would always need these extra items. Maybe we should add as options the the "AddCastleWindsor" to make it easier for people. I will make a PR.

image

image

**Comment by frankieleef** [comment]
Hi, I have been playing around with this new facility and I am experiencing strange problems when deploying a dotnet core application to a Windows VM. Whenever I am running my application on my local machine, I don't experience any errors. But on a Windows VM (hosted on IIS) I am getting a `ComponentNotFoundException`.

`ComponentNotFoundException: No component for supporting the service Namespace.Controllers.ValuesController was found
Castle.MicroKernel.DefaultKernel.Castle.MicroKernel.IKernelInternal.Resolve(Type service, IDictionary arguments, IReleasePolicy policy)
Castle.Facilities.AspNetCore.AspNetCoreMvcExtensions+<>c__DisplayClass0_0.<AddCustomControllerActivation>b__0(ControllerContext context)
Castle.Facilities.AspNetCore.Activators.DelegatingControllerActivator.Create(ControllerContext context)
Microsoft.AspNetCore.Mvc.Controllers.ControllerFactoryProvider+<>c__DisplayClass5_0.<CreateControllerFactory>g__CreateController|0(ControllerContext controllerContext)
Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker.Next(ref State next, ref Scope scope, ref object state, ref bool isCompleted)
Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker+<InvokeInnerFilterAsync>d__14.MoveNext()
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker+<InvokeNextResourceFilter>d__22.MoveNext()
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.Rethrow(ResourceExecutedContext context)
Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.Next(ref State next, ref Scope scope, ref object state, ref bool isCompleted)
Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker+<InvokeFilterPipelineAsync>d__17.MoveNext()
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker+<InvokeAsync>d__15.MoveNext()
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Microsoft.AspNetCore.Builder.RouterMiddleware+<Invoke>d__4.MoveNext()
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware+<Invoke>d__7.MoveNext()`

My `Startup.cs` looks the following:

```
public class Startup
    {
        private readonly WindsorContainer container = new WindsorContainer();
        public Startup(IConfiguration configuration)
        {
            Configuration = configuration;
        }
        public IConfiguration Configuration { get; }

        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvc();
            services.AddCastleWindsor(container);
        }

        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            RegisterApplicationComponents();

            app.UseMvc();
        }

        private void RegisterApplicationComponents()
        {
            container.Install(new ContainerInstaller(Configuration));
        }
    }
```


`ContainerInstaller.cs`

```
 public class ContainerInstaller : IWindsorInstaller
    {
        private Microsoft.Extensions.Configuration.IConfiguration _configuration;

        public ContainerInstaller(Microsoft.Extensions.Configuration.IConfiguration configuration)
        {
            _configuration = configuration;
        }

        public void Install(IWindsorContainer container, IConfigurationStore store)
        {
            container.Register(Component.For<IBaseRepository>().ImplementedBy<MongoDBRepository>().LifestyleTransient());
            container.Register(Component.For<IConfiguration>().ImplementedBy<MongoConfiguration>().LifestyleTransient());

            container.Register(Component.For<NamespaceService>().LifestyleTransient());

            container.Register(Component.For<Namespace.PersonResolver>().LifestyleTransient());
            container.Register(Component.For<Namespace.ChildResolver>().LifestyleTransient());
            container.Register(Component.For<Namespace.HouseResolver>().LifestyleTransient());
            container.Register(Component.For<Namespace.IncomeResolver>().LifestyleTransient());

            container.Register(Component.For<NamespaceResponseConverter>().LifestyleTransient());
            container.Register(Component.For<Namespace.CalculationConverter>().LifestyleTransient());

            container.Register(
                Component.For<IMapper>().UsingFactoryMethod(x =>
                {
                    return new MapperConfiguration(c =>
                    {
                        c.ConstructServicesUsing(container.Resolve);
                        c.AddProfile<Namespace.Mapper>();
                    }).CreateMapper();
                }));
```

I am not sure if this is a Castle Windsor issue, but when hosting it on IIS on my local machine I don't experience the same error. I was hoping one of you would be able to help me out.

fir3pho3nixx:
Any chance you could check it against https://github.com/fir3pho3nixx/Windsor/tree/aspnet-core-windsor-final

I do know we have a memory leak at the moment.

fir3pho3nixx:
@frankieleef please report issues here: https://github.com/fir3pho3nixx/Windsor.AspNetCore/issues

frankieleef:
Let me try that first before creating an issue. Thanks!

PR #377, opened 2018-02-08
Obsolete methods that can be removed safely

fir3pho3nixx:
This addresses obsolescence from issue https://github.com/castleproject/Windsor/issues/338

This PR mainly targets:

 - ActAs
 - Parameters
 - ServiceOverrides 
 - Properties
 - Most methods in DefaultKernel_ObsoleteAPI
 - Most methods in WindsorContainer_ObsoleteAPI

Obsolete methods that can be removed safely:

 - Removed obsolete Resolve methods from IKernel/DefaultKernel && IWindsorContainer/WindsorContainer
 - Removed the obsolete AddComponent methods from IKernel/DefaultKernel
 - Removed the obsolete indexer from IKernel/DefaultKernel && IWindsorContainer/WindsorContainer
 - Removed the obsolete Properties methods from Component Registration
 - Removed the obsolete ServiceOverrides methods from Component Registration
 - Removed the obsolete Parameter method from Component Registration
 - Removed the obsolete ActAs method from Component Registrations and updated documentation

Obsolete methods that cannot be removed safely:

 - IKernel/DefaultKernel AddFacility(String key, IFacility facility)
 - IWindsorContainer/WindsorContainer AddFacility(String key, IFacility facility)

Still need to update CHANGELOG.md but will wait to avoid merge conflict.

Issue #338, opened 2017-09-17
Deprecated APIs for removal in Windsor v5


jonorossi:
- [ ] Logging Facility: `LoggerImplementation` enum and the `loggingApi` XML property
- [ ] ...

fir3pho3nixx:
Logging Facility -> #341

jonorossi:
As I wrote in #341 this wasn't a task to start doing the work yet, but tagged "discussion" so we can plan what we want to remove, we haven't even released those methods being marked `Obsolete` yet.

fir3pho3nixx:
I do believe most of the XML configuration can be removed. I think XML removes the compiler from type safety resolution and promotes these problems to the runtime environment. So I would think any XML based config needs to hit the road and go to that XML heaven in the sky.

fir3pho3nixx:
I also think parent/child containers should be removed.

fir3pho3nixx:
It also seems like we should get rid of parent/child containers. ref:https://github.com/castleproject/Windsor/issues/115

fir3pho3nixx:
ComponentRegistration could do with some sprucing up. 

- [ActAs](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/ComponentRegistration.cs#L145)

- [Parameters](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/ComponentRegistration.cs#L889)

- [ServiceOverrides](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/ComponentRegistration.cs#L928)

- [Properties](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/ComponentRegistration.cs#L1173)

fir3pho3nixx:
BasedOnDescriptor has a few obsoletes also.

- [BasedOn'T](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs#L70)
- [Where](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/BasedOnDescriptor.cs#L196)

fir3pho3nixx:
[...]

PR #378, opened 2018-02-09
Obsolete methods that can be removed with bigger impact

fir3pho3nixx:
To support the deprecation of IKernel/DefaultKernel AddFacility(String key, IFacility facility) I also removed IConfiguration from IFacility.Init interface.

This is the last piece of the puzzle when it comes to getting rid of obsolete interface definitions on IKernel/IWindsorContainer. 

For low hanging fruit review [this PR first](https://github.com/castleproject/Windsor/pull/377).


fir3pho3nixx:
The docs for this definitely need updating will do that after we have reviewed the code.

jonorossi:
> To support the deprecation of IKernel/DefaultKernel AddFacility(String key, IFacility facility) I also removed IConfiguration from IFacility.Init interface.

Is there a reason you removed `IConfiguration` from facilities as part of this pull request, I thought it was just going to be getting rid of a bunch of obsolete `AddFacility` methods and moving a bit of code around as they mostly called each other.

I think we need to be much clearer with users if we are deprecating or removing support for XML configuration for any aspect of Windsor, XML configuration has been a big feature in Windsor's history with plenty of use out there, so I think it must be explicitly communicated with its own issue in the issue tracker even if that is just removing XML configuration support for facilities but it stays for installers (for example).

fir3pho3nixx:
> Is there a reason you removed IConfiguration from facilities as part of this pull request, I thought it was just going to be getting rid of a bunch of obsolete AddFacility methods and moving a bit of code around as they mostly called each other.

When looking at the [AddFacility methods in the DefaultKernel_ObsoleteAPI](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/DefaultKernel_ObsoleteAPI.cs#L25) I first tried just deleting them from the project, but then I noticed a compiler error because one of those are being used by the [AddFacility method](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/DefaultKernel.cs#L279) in the DefaultKernel with pragma flags. This got me round to thinking about the obsolete method (which at this point really did not feel so obsolete anymore). So I then reworked the method to get rid of the `string key`/`ConfigurationStore` dependency and put that implementation into the DefaultKernel. This got me round to thinking about how IFacility.Init is being called, because if we are deprecating the `string key` parameter from the public API then perhaps the intention here was also to get rid of IConfiguration from the IFacility.Init interface too? This is where I became a little unsure and thought putting through an experimental PR would be the right thing to do. 

Based on your reaction, this is a cut too deep for how facilities work :)

What I can do is hide the obsolete method by making it internal and then just keep the convention of `string key` to mean looking up a configuration based on the facility type fullname. 

What do you think?

fir3pho3nixx:
> I think we need to be much clearer with users if we are deprecating or removing support for XML configuration for any aspect of Windsor, XML configuration has been a big feature in Windsor's history with plenty of use out there, so I think it must be explicitly communicated with its own issue in the issue tracker even if that is just removing XML configuration support for facilities but it stays for installers (for example).

Most of the deprecation activities on #338 might be to big for v5 too, we can close that out and raise new issues for the items we discussed there. That said, there is still plenty of Obsolescence that could be cleaned up.

jonorossi:
I'm not exactly clear what you are proposing, would you mind updating the pull request with what you are proposing.

I think that XML configuration is still quite valuable for facilities, many people have facilities not shipped in the box where they wire up the facility by code but much of the facility is configured by external configuration depending on where/how it is deployed.

fir3pho3nixx:
> I'm not exactly clear what you are proposing, would you mind updating the pull request with what you are proposing.

I have made the change I was talking about. Code reads better than English! :) 

I think this can now be classified as a safe change. XML config preserved 👍

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -278,9 +279,22 @@ IHandler IKernelInternal.CreateHandler(ComponentModel model)
...
+			var facilityType = facility.GetType();
+			if (facilities.Any(f => f.GetType() == facilityType))
+			{
+				throw new ArgumentException(
+					string.Format(
+						"Facility of type '{0}' has already been registered with the container. Only one facility of a given type can exist in the container.",
+						facilityType.FullName));
+			}
+			facilities.Add(facility);
+			facility.Init(this, ConfigurationStore.GetFacilityConfiguration(facility != null ? facility.GetType().FullName : null));
```
jonorossi: `facility` can't be null here.

At first this appears to be a breaking change to loading of facility configuration from an XML file because you've removed the ability to define the "id" attribute which both the code and XML share, however that breaking change was actually made in 3.0.0 according to our changelog:

> change - type attribute is now required and id is ignored in facility XML configuration
> impact - low
> fixability - easy
> description - Since type is uniquely identifying facilities there was no point in keeping the id
> 	around anymore.
> fix - This change can affect you in two ways. If you were using facilities node in the XML and
> 	not specifying the type it is now mandatory. Notice Windsor's ability to apply short type
> 	names works here as well, so often just type name is enough - no need to specify assembly
> 	qualified name. Also the assembly will now be instantiated by the container, so if you were
> 	adding it in code later on, this is no longer required (in fact it will throw an exception
> 	saying the assembly was already added).
> 	The other thing that may affect you is if you were looking up facility config namnually via
> 	IConfigurationStore.GetFacilityConfiguration method. It now expects full name of the type
> 	as the key, so you should be calling it like this:
> 	store.GetFacilityConfiguration(typeof(YourFacility).FullName);

What we should actually do here is remove any mention of the `id` attribute in our docs, XML docs and unit tests since that attribute is ignored, `DefaultConfigurationStore.facilities` holds a map of type full name to facility instance.

`XmlInterpreter ` and `DefaultComponentInstaller` definitely ignore `id`, I'm thinking that it should actually throw on the existence of the attribute now that so much time has passed and because it just confuses people that the `id` attribute is still actually used.

What do you think?

fir3pho3nixx: Yup, dealt with. I updated it everywhere.

docs/atm-facility.md
```diff
@@ -40,7 +40,6 @@ public class BusinessClass
 <configuration>
    <components>
       <component
-         id="mycomp"
```
fir3pho3nixx: Whoops! This needs to be put back.

fir3pho3nixx:
Let me know if I can squish.

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
jonorossi: Thanks.

Did you have an opinion on my last part of that comment, throwing if someone does actually specify an `id` attribute to force people to remove them when they upgrade to clean up their XML configuration? Happy if you think it is more work than it is worth, especially that you've just removed the obsolete `AddFacility` methods so users should realise the id/key is ignored if they were using those ones, otherwise there is no harm.

fir3pho3nixx:
I completely forgot about the exception. Let me add it before we merge.

fir3pho3nixx:
@jonorossi - I added a test which asserts that any facility config with an 'id' attribute will now throw a ConfigurationProcessingException and squished down the commits.

jonorossi:
Looks great, thanks.

PR #383, opened 2018-02-28
Add UseCastleWindsor(this IApplicationBuilder app, IWindsorContainer …

generik0:
Add UseCastleWindsor(this IApplicationBuilder app, IWindsorContainer container) 
Make tests test both with and without type for assembly

#379


src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -53,9 +54,26 @@ public static void AddCastleWindsor(this IServiceCollection services, IWindsorCo
...
+			var assembly = Assembly.GetEntryAssembly();
+			UseCastleWindsor(container, assembly);
+		}
+
+		private static void UseCastleWindsor(IWindsorContainer container, Assembly assembly)
+		{
+			container.Register(Classes.FromAssemblyInThisApplication(assembly).BasedOn<Controller>().LifestyleScoped());
+			container.Register(Classes.FromAssemblyInThisApplication(assembly).BasedOn<ViewComponent>().LifestyleTransient());
+			container.Register(Classes.FromAssemblyInThisApplication(assembly).BasedOn<TagHelper>().LifestyleTransient());
 		}
```
fir3pho3nixx: ^^ This method should probably be moved to the bottom of the file because it is private. 

https://github.com/castleproject/Home/blob/master/coding-standards.md#type-members

fir3pho3nixx:
@generik0 - Just tested this out and it works fine. Thanks.

@jonorossi - Could you give it a look as per the comments [here](https://github.com/castleproject/Windsor/pull/376#issuecomment-369006527)?

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -53,9 +54,26 @@ public static void AddCastleWindsor(this IServiceCollection services, IWindsorCo
...
+		}
+
+		/// <summary>
+		/// Use this to register all Controllers, ViewComponents and TagHelpers. Will register the entry assembly and its referenced items
+		/// </summary>
+		/// <param name="app">Application builder retained as extension</param>
+		/// <param name="container">Windsor container to register framework types in</param>
+		public static void UseCastleWindsor(this IApplicationBuilder app, IWindsorContainer container)
+		{
+			var assembly = Assembly.GetEntryAssembly();
```
jonorossi: I would have thought this would use GetCallingAssembly (marked with no inlining) like Windsor's fluent registration API, not GetEntryAssembly.

This project in the solution is built against .NET Standard 2.0 not 1.6 so the method is back.

generik0: Thanks for your reply :-)
I have previously tested a lot with entry vs calling. Entry is the correct one.
Then e.g. the project with the bootstrapped referencing this package can be in another project than the startup.
The point of the method to me. Register for the entry/startup assembly / process. The entry assembly is the culmination of everything and the top level. Thoughts?

jonorossi: I'm happy to take yours and @fir3pho3nixx lead, I just wanted to make sure it was a fully conscious decision.

generik0: I perfectly agree. I am happy for decision. Comments are welcome.
I will change if you guys agree it should be different.
I have had many issues with the get assembly calls. But recently most luck with aspnetcore and entry.
Any other experiences?

fir3pho3nixx: I think this should be fine for now. I tested this with a published app and in visual studio and it worked.

fir3pho3nixx:
@generik0 can you please move the private method to the bottom of the class as per the Castle Coding guidelines and I think we are good to merge.

generik0:
I will ASAP 
I have ben thinking about a feature both Autofac and SimpleInjector do not have for AspNetCore.
The app config files are now in Json and the injection of config parameters is different than before.
I would like to add a resolver for the aspnetcore Json settings. I will make a PR when I have investigated more :-)

fir3pho3nixx:
@generik0 thanks. Sorry I have not been so responsive, I am in Japan for the next two weeks. Please raise a new issue so we can discuss some of your idea’s. Might be able to help research some of the things you are thinking of and we can work together to formalise it better before we release. Thanks for contributing, really appreciate it ;)

generik0:
@fir3pho3nixx i also moved the test setup as it is no longer a [Setup]

fir3pho3nixx:
@generik0 looks like we have a merge conflict. Can you fix it? I might have stepped on your toes with a previous commit. Sorry!

generik0:
I wil rebase with Master :-)

jonorossi:
Thanks

PR #384, opened 2018-03-07
Removing deprecated AllTypes and AllTypesOf

fir3pho3nixx:
This PR addresses the removal of AllTypes and AllTypesOf which have been flagged as obsolete based on issue #338


jonorossi:
That's an easy one, no unit tests where even using them 👍

PR #385, opened 2018-03-08
Removing deprecated BasedOn methods

fir3pho3nixx:
This PR addresses the removal of BasedOn methods that reset component registrations when fluently chained from issue #338.


jonorossi:
That's a bit of a strange unit test the one you modified, but all good.

fir3pho3nixx:
I agree, the unit test is a little weird. It now asserts that if you call the same method repeatedly, that the BasedOnDescriptor still works. Without background on the original problem I could see why the obsolescence was required. The test does still exercise the fluent `if` branch for predicates. So rather than delete it I thought it would be good if could hang around for a bit.

fir3pho3nixx:
Happy to do more work in this area, just need a few tips on where we go from here.

PR #387, opened 2018-03-18
Add exception when directly registering a AssemblyDescriptor object

### LINKED ISSUE #374: Registering Classes.FromThisAssembly() doesn't do anything
Trying to register all classes with the following syntax does nothing, although it compiles just fine and there's nothing obvious to indicate why it wouldn't register all types in my assembly.

```c#
container.Register(Classes.FromThisAssembly());
```

The *correct* syntax, I've finally discovered, is to use the `.Pick()` descriptor

```c#
container.Register(Classes.FromThisAssembly().Pick());
```
It seems to me that that the registration should either default to use Pick() if no other descriptor is provided or throw an exception indicating the operation isn't allowed. A better solution would be, if possible, to prevent passing in the return type of any `FromAssembly())` method into the registration method, although I suspect that would be fairly more complicated.

danquist:
Resolves #374

### DIFF
--- a/src/Castle.Windsor.Tests/Registration/AllTypesTestCase.cs
+++ b/src/Castle.Windsor.Tests/Registration/AllTypesTestCase.cs
@@ -617,5 +617,15 @@ namespace CastleTests.Registration
 
 			Assert.AreEqual(1, services.Length);
 		}
+
+		[Test]
+		public void RegisterGenericTypes_WithoutDescriptor()
+		{
+			var ex = Assert.Throws<ArgumentException>(() => {
+				Kernel.Register(Classes
+					.FromAssembly(GetCurrentAssembly())
+				);
+			});
+		}
 	}
 }
--- a/src/Castle.Windsor/MicroKernel/DefaultKernel.cs
+++ b/src/Castle.Windsor/MicroKernel/DefaultKernel.cs
@@ -537,6 +537,10 @@ namespace Castle.MicroKernel
 			var token = OptimizeDependencyResolution();
 			foreach (var registration in registrations)
 			{
+				if (registration is FromAssemblyDescriptor)
+				{
+					throw new ArgumentException("Assembly registrations must be appended with a BasedOnDescriptor.\nUse '.Pick()' to select all types in the assembly.");
+				}
 				registration.Register(this);
 			}
 			if (token != null)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Enhancements:
 - Added XML documentation to BeginScope and RequireScope lifetime extensions (@jonorossi)
 - Upgraded build to use NUnit Adapters (@fir3pho3nixx, #243)
+- Added ArgumentException thrown when attempting to directly register a AssemblyDescriptor object
 
 Breaking changes:
 - Built-in System.Web support has been moved to the new Castle.Facilities.AspNet.SystemWeb facility (@fir3pho3nixx, #283)


fir3pho3nixx:
@danquist - Could you update the CHANGELOG.md with the details of the change?

danquist:
Ah, yes sure thing.

danquist:
@fir3pho3nixx would this be considered a breaking change? The old call never did anything, but if someone has this call anywhere, it will now throw an exception whereas it wouldn't before.

fir3pho3nixx:
Given the original method did nothing, I would say this is an enhancement, improved UX. We are going to increment the major version number anyway as a result of all the deprecation work I have been doing.  Either way you are covered! :)

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -537,6 +537,10 @@ public IKernel Register(params IRegistration[] registrations)
 			var token = OptimizeDependencyResolution();
 			foreach (var registration in registrations)
 			{
+				if (registration is FromAssemblyDescriptor)
+				{
+					throw new ArgumentException("Assembly registrations must be appended with a BasedOnDescriptor.\nUse '.Pick()' to select all types in the assembly.");
```
fir3pho3nixx: Complete nitpick. Can we improve this exception to specify the public API's for component registrations as opposed to the fluent registration types?

You know something like: 

"It looks like you tried to register all types in an assembly using `container.Register(Classes.FromThisAssembly());`. It is recommended that you at least use some kind of criteria like `container.Register(Classes.FromThisAssembly().BasedOn<IFoo>());`. If you meant to register everything, please use `container.Register(Classes.FromThisAssembly().Pick());` instead."

src/Castle.Windsor.Tests/Registration/AllTypesTestCase.cs
```diff
@@ -617,5 +617,15 @@ public void RegisterGenericTypes_MultipleBasedOnWithTheSameTypesTwice_SelectedAs
...
+
+		[Test]
+		public void RegisterGenericTypes_WithoutDescriptor()
+		{
+			var ex = Assert.Throws<ArgumentException>(() => {
+				Kernel.Register(Classes
+					.FromAssembly(GetCurrentAssembly())
+				);
+			});
+		}
```
fir3pho3nixx: Do we need to assign `var ex` if we are not making any assertions on the message returned?

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -537,6 +537,10 @@ public IKernel Register(params IRegistration[] registrations)
 			var token = OptimizeDependencyResolution();
 			foreach (var registration in registrations)
 			{
+				if (registration is FromAssemblyDescriptor)
```
jonorossi: Sorry for not getting to review this sooner.

Is there a reason you put this in the kernel rather than in the descriptor itself like I suggested in the issue?

https://github.com/castleproject/Windsor/blob/c192ed68f3b8ea22f0cd67f3922c9029a5de58e2/src/Castle.Windsor/MicroKernel/Registration/FromDescriptor.cs#L173-L178

If you didn't want the exception to also apply to `FromTypesDescriptor` (which has the same type of problem), then I think it would be best to override `Register` in `FromAssemblyDescriptor` and perform the check there before calling `base`.

fir3pho3nixx:
@danquist - I am happy to submit a PR on your fork if you need help. Hope you don't see this as a bashing of some malevolent kind. Not what we meant fella.

fir3pho3nixx:
@danquist, @jonorossi I would like to keep PR's current. Anything older than 3 months is outta here. 

Please re-open if you disagree.

PR #389, opened 2018-03-27
Add registration options - PR just for discussion

generik0:
Hi

This PR is just for discussion.
Would it not be a good idea to add an option possibility to add for extra's?

### DIFF
--- a/src/Castle.Facilities.AspNetCore.Tests/AspNetCoreFacilityTestCase.cs
+++ b/src/Castle.Facilities.AspNetCore.Tests/AspNetCoreFacilityTestCase.cs
@@ -99,7 +99,13 @@ namespace Castle.Facilities.AspNetCore.Tests
 		private void BuildWindsorContainer(ServiceCollection serviceCollection)
 		{
 			container = new WindsorContainer();
-			serviceCollection.AddCastleWindsor(container);
+			serviceCollection.AddCastleWindsor(container,
+				options =>
+				{
+					options.AddHttpContextAccessor();
+					options.AddContainerResolvedScoped<IFakeFrameworkComponent>();
+					options.AddContainerResolvedScoped<IFakeFrameworkComponentEnumerable>();
+				});
 			container.Register(Component.For<AnyService>());
 		}
 
--- a/src/Castle.Facilities.AspNetCore/Options/RegistrationOptions.cs
+++ b/src/Castle.Facilities.AspNetCore/Options/RegistrationOptions.cs
@@ -1,0 +1,56 @@
+﻿// Copyright 2004-2018 Castle Project - http://www.castleproject.org/
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.Facilities.AspNetCore.Options
+{
+	using System;
+
+	using Castle.Windsor;
+
+	using Microsoft.AspNetCore.Http;
+	using Microsoft.Extensions.DependencyInjection;
+
+	public class RegistrationOptions
+	{
+
+		private readonly IServiceCollection services;
+		private readonly IWindsorContainer container;
+
+		public RegistrationOptions(IServiceCollection services, IWindsorContainer container)
+		{
+			this.services = services;
+			this.container = container;
+		}
+
+		public void AddHttpContextAccessor()
+		{
+			services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+		}
+
+		public void AddContainerResolvedScoped<T>() where T : class
+		{
+			services.AddScoped(provider => container.Resolve<T>());
+		}
+
+		public void AddContainerResolvedSingleton<T>() where T : class
+		{
+			services.AddSingleton(provider => container.Resolve<T>());
+		}
+
+		public void AddContainerResolvedTransient<T>() where T : class
+		{
+			services.AddTransient (provider => container.Resolve<T>());
+		}
+	}
+}
--- a/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
+++ b/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
@@ -18,6 +18,7 @@ namespace Castle.Facilities.AspNetCore
 	using System.Linq;
 	using System.Reflection;
 
+	using Castle.Facilities.AspNetCore.Options;
 	using Castle.Facilities.AspNetCore.Resolvers;
 	using Castle.MicroKernel.Lifestyle;
 	using Castle.MicroKernel.Registration;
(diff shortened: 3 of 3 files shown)


src/Castle.Facilities.AspNetCore.Tests/AspNetCoreFacilityTestCase.cs
```diff
@@ -99,7 +99,13 @@ private static ServiceCollection BuildServiceCollection()
...
 		{
 			container = new WindsorContainer();
-			serviceCollection.AddCastleWindsor(container);
+			serviceCollection.AddCastleWindsor(container,
+				options =>
+				{
+					options.AddHttpContextAccessor();
+					options.AddContainerResolvedScoped<IFakeFrameworkComponent>();
+					options.AddContainerResolvedScoped<IFakeFrameworkComponentEnumerable>();
+				});
```
fir3pho3nixx: Why can't this be done on the original service collection? Those API's are well documented. All the third party extensions would eventually need to integrate with these API's. What problem this trying to solve?

src/Castle.Facilities.AspNetCore/Options/RegistrationOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		public void AddHttpContextAccessor()
+		{
+			services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+		}
+
+		public void AddContainerResolvedScoped<T>() where T : class
+		{
+			services.AddScoped(provider => container.Resolve<T>());
```
fir3pho3nixx: At the moment, we have activators which call out of ASPNET Core and resolve from the Windsor Container. The Container itself then uses a sub dependency resolver which resolves framework components from the IServiceCollection by building a service provider. 

By adding this code you are trying to make dependency resolution flow both ways between the service provider and the container. I imagine situations like infinite resolution cycles because the containers keep resolving from each other. It could be hard to maintain this approach.

generik0:
Hi @fir3pho3nixx. No stress...

Reg: AddHttpContextAccessor
If you don't add the httpConext registration this way, it does not appear to work with the facility.
Your facility does not register the IHttpContextAccessor nor does this work:
container.Register(Component.For<IHttpContextAccessor>()
                .ImplementedBy<HttpContextAccessor>());
Hence it needst to be manually registered to the service.
There are other examples of this, like for languaging...


Reg: services.AddScoped(provider => container.Resolve<T>());
if you do not add this, then the @inject does not work.
This is all resolving the issue that we have already talked about.  The link you sent me had this as an example / recommended solution to getting @inject to work...

I..e some things need to be registered to the serrvic collection, as the facility cannot resolver them even though they are registerd...

Please look at the @inject simple injector issue link you sent me :-)

src/Castle.Facilities.AspNetCore.Tests/AspNetCoreFacilityTestCase.cs
generik0: I am just adding into the facity to add extra options for the castle registation so everythng to the service collection is nicely wrapped.
I think it is very powerful and ver neet. Only the facility needst to know how to work with the service collection :-)

fir3pho3nixx: This is dangerous and could have unintended consequences. Wrapping the IHttpContextAccessor could mean we are taking a dependency on a type which has been flagged for deprecation. Please see: https://github.com/aspnet/Announcements/issues/190

generik0:
@fir3pho3nixx Thanks for your answer.
regarding the injector, i read your link to simple injector issue, and så the container resolver.
Personally i dislike the static service locator very very much. It reminds me og webforms.

Like the simple injector dudes, i prefer the container resolving.
We only ever use 1 inject. a CqrsDispatcher, so we have 1 @inject for "everything".

generik0:
@fir3pho3nixx regarding the httpaccessor, i might not have the lates changes from your faciltiy that have the accesor. I will update.

I understand your point about opening a can of worms here. Fair enough.
I will continue to manually register my languagning and cqrs for injection :-)

fir3pho3nixx:
> @fir3pho3nixx Thanks for your answer.

No problem. I like the fact that we are sense checking this together. 👍 

> Personally i dislike the static service locator very very much. It reminds me og webforms.

Ha! Can see how bad that could be :) It is an old style but it still works. You practically get a 1-to-1 feature mapping. The syntax is uglier but it definitely does the job of @Inject without the hassle of 2 registrations. I do believe we need to update docs for this at the very minimum. I will pick this up once I have finished the v5 deprecation work. 

> Like the simple injector dudes, i prefer the container resolving.

Me too.

> We only ever use 1 inject. a CqrsDispatcher, so we have 1 @Inject for "everything".

You are very lucky you chose this pattern. A single `double registration` does not sound that bad. If I was working in a project where everything is a service than I think it would very quickly grow out of control.  

> I will continue to manually register my languagning and cqrs for injection :-)

Can you share what you are doing for the multi-lingual stuff?

generik0:
Sure.
To use IHtmlLocalizer<SharedResource> as @inject and in razor this needs to be registered:
image

So this comes back to the to make the double registration or not.
Basically, i will continue to do the double registration for this IHtmlLocalizer and for the cqrsdispatcher :-)

generik0:
Hi @fir3pho3nixx 
I have been thinking about it more and more. I do feel that it is a good solution when we absolutly must "cross-wire".
We could even call the options, crossWiringOptions.

I do believe people will need these options for the registration, lets make a framework that it easy for them. To stop the double registration we could even make the framework within the options actions:

1. Register the component as fallback in castle
2. Register with service collector.

I really think we should make this options/wrapper like you simpleinjector link points out. https://github.com/simpleinjector/SimpleInjector/issues/362 regarding cross-wiring.
Only make the facility help it be done. neatly for people...

fir3pho3nixx:
Hi @generik0 

```
I do believe people will need these options for the registration, lets make a framework that it easy for them. To stop the double registration we could even make the framework within the options actions:

Register the component as fallback in castle
Register with service collector.
```
Based on this we are going to push for a 2 way resolution mechanism between Windsor and the ServiceProvider. We need to be careful of "Torn Lifestyles", for example where both containers have the same service registered as a singleton and they both provide independent instances during a request. 

Looks like we are wrangling with 2 distinct features here:

## 1. [Controller | TagHelper | ViewComponent]Activator *resolves from* Windsor *which should resolve framework components from* ServiceProvider [via] SubResolver.

After looking through the extension which enables cross wiring over [here](https://github.com/simpleinjector/SimpleInjector/blob/master/src/SimpleInjector.Integration.AspNetCore/SimpleInjectorAspNetIntegrationExtensions.cs#L247) it appears that they backing [unresolved types for SimpleInjector] onto the [ServiceProvider] (https://github.com/simpleinjector/SimpleInjector/blob/master/src/SimpleInjector.Integration.AspNetCore/SimpleInjectorAspNetIntegrationExtensions.cs#L301) for Microsoft Dependency Injections. Have we not already achieved this with our [sub resolver](https://github.com/fir3pho3nixx/Windsor/blob/0ac809f8fe68be6e5d99bde6224a0df2f846a945/src/Castle.Facilities.AspNetCore/Resolvers/FrameworkConfigurationDependencyResolver.cs)? There is already cross wiring from Windsor -> ServiceProvider.

## 2. @Inject *resolves from* ServiceProvider *which cannot resolve from* Windsor.

You have already proven you can get this to work using a late bound registration in the IServiceCollection which calls WindsorContainer.Resolve. If all we need is late bound factory methods for this to work, wouldn't it be easier to subscribe to the [MicroKernel.ComponentRegistered](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/DefaultKernel_Events.cs#L132) event and use the ComponentModel argument to create a registration in the IServiceCollection like you did in this PR: services.AddScoped(provider => container.Resolve<T>());?

This would potentially meet both our requirements of:

1. Your requirement of not needing to register things twice.
2. My requirement of avoiding creating a 3rd flavour of registration API.

What do you think?

generik0:
@fir3pho3nixx, Dude (if you don’t mind me calling you that), you really know your shit. It a pleasure to work with you.
If you have a better idea to getting the @inject to work using microkernel let us give it a go. I have never been so deep in castle. Everything else is working with the facility! I am so happy to get away from autofac (even though Autofac is also ok).

My main point is just there should be a nice option set to do special stuff :-)
Will we get any tracking issues with these resolved objects. I know my factory objects need to be released. Will castle handle this for these resolved items?

Should I make a prototype with the micro kernel or will you? I might need a like to an example. This is very exciting :-)

fir3pho3nixx:
I am looking at this right now. Will let you know once I have it working :)

fir3pho3nixx:
I have this working, give me a few minutes to refactor it a little.

fir3pho3nixx:
Ok, I have some code up.

I removed the validation method against aspnet core framework components being installed in Windsor, as we really do not have that problem anymore. Now if you install the HttpContextAccessor in Windsor it will automatically install a late bound component in the IServiceCollection which asks for the instance from Windsor during resolve time. @Inject vualla ... working :)

The new Startup.cs is [here](https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/context-accessor/WebApp/Startup.cs)

The new facility code is [here](https://github.com/fir3pho3nixx/Windsor/blob/3056f4758640818805f169665a1d0178a4f41f90/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs)

The working @inject example is now [here](https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/context-accessor/WebApp/Views/Home/Index.cshtml#L10)

I will raise a proper PR and explain to @jonorossi why we needed this change. Thanks for poking me on this @generik0, I think we ended up in a good place :)

fir3pho3nixx:
> Will we get any tracking issues with these resolved objects. I know my factory objects need to be released. Will castle handle this for these resolved items?

I think you have stumbled on to another flaw we might need to address. Anything resolved from Windsor should be dealt with. Notice how the controller activator makes releasing optional [here](https://github.com/fir3pho3nixx/Windsor/blob/3056f4758640818805f169665a1d0178a4f41f90/src/Castle.Facilities.AspNetCore/Activators/DelegatingControllerActivator.cs#L30)?

We should implement this in all the activators if we can, and make sure this not optional so that Container.Release is always called for TagHelpers, ViewComponents and Controllers. Singletons are luckily not a problem. Scoped is taken care of by the auto scoping startup filter, I am not entirely convinced the transients for framework things like the IHttpContextAccessor are safe and we would need to check things like this for memory leaks if they are not disposed.

fir3pho3nixx:
I will add this in when I raise the PR tomorrow.

generik0:
@fir3pho3nixx 
Very cool!
I will grab your fork, build and ingratiate it tomorrow. I will let you know what I experience:-)

PR #390, opened 2018-03-27
Fix name of method in documentation header

stwalkerster:
IInitializable has the method Initialize, not Initialized


jonorossi:
Thanks @stwalkerster, merged.

PR #391, opened 2018-03-31
Removing deprecated NoTrackingReleasePolicy

fir3pho3nixx:
This gets rid of the obsolete NoTrackingReleasePolicy. 

@jonorossi this was an interesting one. I was not sure if the intent behind this was to get rid of the feature altogether. I opted for caution here and retained this functionality by simply allowing the setting of Container.Kernel.ReleasePolicy to null. 

#338


hconceicao:
This class is/was the recommended way to go to get rid of the mem leak caused by component burden when mixing transient and singleton components (for example: controllers). 

Although the feature is still there, as @fir3pho3nixx mentioned, people probably will have a hard time to adapt.

fir3pho3nixx:
Thank you Henry. I have noticed that under certain conditions burdens do not honour singletons because they are not lifecycle aware. This is causing untold pain when it comes to typed factories. Crikey, how do we fix this? Can we close this PR and do it properly? Would need some help crafting the issue with explicit test cases. Happy to do the work.

fir3pho3nixx:
Honestly I want to rework burdens and typed factories to be smarter.

jonorossi:
> what was the point of getting rid of this class? Do you know?

I don't think this was marked obsolete in the same spirit as other types/members as Henry also mentioned, it was marked obsolete to give people a compiler warning that this is an ugly workaround that we don't have an easy solution for at the moment.

I think it would be better to keep the class and not implicitly support this scenario with a `null` release policy, that just hides the problem with something that looks valid.

fir3pho3nixx:
@jonorossi - I agree. I think we should leave this one out until I understand the cause of the memory leaks better.

fir3pho3nixx:
@hconceicao - Do you know where I can gain more info on how to re-produce the memory leak?

PR #394, opened 2018-04-15
ASP.NET Core Facility (now production ready BUT... )

fir3pho3nixx:
This is the final PR for the ASP.NET Core Facility. The previous design did not support the @Inject Razor directive. The problem was a difficult one of timing registrations in the right places and introducing the notion of `Cross Wiring`. 

`Cross wiring` is provided as an extension method to component registration which create's late bound registrations in the IServiceCollection(which are basically factories) that are backed on to a Windsor Container.Resolve. This removes the burden of Castle Windsor users of having to maintain 2 sets of registration logic and moves them toward the Windsor registration API. It also makes things like the @Inject Razor directive work. 

@generik0 and I also checked to make sure that we have fixed all the problems with generics, I think there was a gremlin in the previous implementation. It got magically fixed when I re-wrote the generic implementation with additional unit tests.

This code has also endured a significant bashing to make sure it does not have any memory leaks by using the performance tests in this [repo](https://github.com/fir3pho3nixx/Windsor.AspNetCore). I am satisfied that for scoped lifestyles(Controller, ViewComponents, TagHelpers and cross wired @Inject consumers in the view) there are no memory leaks(roughly 300MB foot print) and it serves 250+ requests/second. This is great for anyone using this in docker.

We need to start thinking about releasing V5. @jonorossi did you have a timescale in mind?


generik0:
Hi @jonorossi, not to be a stresshead, but could you please review this PR ASAP. 
@fir3pho3nixx has done a great job, and we really did spaend a good amount of time discussing various matters. It has ended in a good result. Maybe even one of the cleanest integrations to aspnetcore service collector compared to all..
If it can be reviewed, any issues found fixe, and merged, that would be really, really great.. Cheers...

generik0:
@fir3pho3nixx I found one thing. Maybe for the future?
But i can not add any configurations for the registration of RegisterCrossWired, like e.g. the "Named" nor "IsDefault". Maybe there needst to be added a parameter to RegisterCrossWired or an extension for CastleWindsorRegistrationOptions that allows this?

Also, maybe CastleWindsorRegistrationOptions  should be renamed CastleWindsorAspNetCoreFacilityRegistrationOptions?

fir3pho3nixx:
Hi @generik0 

> But i can not add any configurations for the registration of RegisterCrossWired, like e.g. the "Named" nor "IsDefault". Maybe there needs to be added a parameter to RegisterCrossWired or an extension for CastleWindsorRegistrationOptions that allows this?

There is already one. [Here](https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs#L229) and [here](https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs#L282).

[Example](https://github.com/fir3pho3nixx/Windsor.AspNetCore/blob/master/WebApp/Startup.cs#L83): 

```csharp
private void RegisterApplicationComponents(IServiceCollection services)
{
	// Cross wiring example, exposes types to Razor @Inject directive
	container.Register(Component.For<IUserService>().ImplementedBy<AspNetUserService>().CrossWired(container, services, LifestyleType.Scoped));
}
```

fir3pho3nixx:
@generik0 I did not want the burden of maintaining a 3rd registration API. I really believe we should delete the [Register](https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs#L132) and [RegisterCrossWired](https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs#L169) methods from the [CastleWindsorRegistrationOptions](https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs#L35) class.

generik0:
@fir3pho3nixx fair enough :-)
So registration is purely? If so, looks good to me, then people will just have to use registration with CrossWired extensions? (as shown in your example):
```c#
container.Register(Component.For<AnyService>().CrossWired(container, serviceCollection, LifestyleType.Singleton));
```
if so, I understand your point...

fir3pho3nixx:
> So registration is purely used, not options?

I made a statement [here](https://github.com/castleproject/Windsor/issues/120#issuecomment-334867589). I intend to honour the original registration API. 

 > If so, looks good to me

Great removing them :)

generik0:
@fir3pho3nixx I do have to admit, I still wished i could register the cross wired without the need to do a component register.  I have had to change some of my fluent registrations to allow for the for  CrossWired services. It was no biggy, just had to be done.

But lets leave it for now. Maybe i am the only one interested in registrering this way...
This is just a footnote.

docs/aspnetcore-facility.md
```diff
@@ -51,7 +91,7 @@ Here is a complete example:
 ```csharp
 public class Startup
 {
-	private readonly WindsorContainer container = new WindsorContainer();
+	public static readonly WindsorContainer container = new WindsorContainer();
```
jonorossi: Is having this public in the example a good idea?

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
```diff
@@ -17,6 +17,10 @@
 		<RootNamespace>Castle.Facilities.AspNetCore</RootNamespace>
 	</PropertyGroup>
 
+	<ItemGroup>
+	  <None Remove="Resolvers\FrameworkConfigurationDependencyResolver.cs~RF4e009575.TMP" />
```
jonorossi: Looks like some junk added to the project file.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -29,44 +29,185 @@ namespace Castle.Facilities.AspNetCore
...
-			services.AddCustomControllerActivation(container.Resolve);
-			services.AddCustomTagHelperActivation(container.Resolve);
-			services.AddCustomViewComponentActivation(container.Resolve);
-			container.Kernel.Resolver.AddSubResolver(new LoggerDependencyResolver(services));
-			container.Kernel.Resolver.AddSubResolver(new FrameworkConfigurationDependencyResolver(services));
+			AddFrameworkResolversToWindsor(services, container);
+			InstallFrameworkIntegration(services, container);
+
+			var options = new CastleWindsorRegistrationOptions(container, services);
+			(configure ?? ((o) => { }))(options);
```
jonorossi: This reads really poorly and would invoke that lambda even though it does nothing, could you change this to an if statement with a null check.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -29,44 +29,185 @@ namespace Castle.Facilities.AspNetCore
...
+		/// <param name="registration">The component registration that gets copied across to the IServiceCollection</param>
+		/// <param name="serviceType">The type of the service</param>
+		/// <param name="container">Windsor container which 'activators'/'cross wired' components resolve from</param>
+		/// <param name="serviceCollection">ASP.NET Core service collection from Microsoft.Extensions.DependencyInjection</param>
+		/// <param name="lifestyleType">The lifestyle type that needs to be copied across, only Singleton, Scoped and Transient are supported</param>
+		public static ComponentRegistration<TService> CrossWired<TService>(this ComponentRegistration<TService> registration, Type serviceType, IWindsorContainer container, IServiceCollection serviceCollection, LifestyleType lifestyleType = LifestyleType.Scoped) where TService : class
 		{
-			var assembly = typeof(TTypeAssembly).Assembly;
-			UseCastleWindsor(container, assembly);
+			registration.LifeStyle.Is(lifestyleType);
```
jonorossi: The comment says only 3 lifestyle types are supported, should this check nothing else is passed in?

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -29,44 +29,185 @@ namespace Castle.Facilities.AspNetCore
...
+			if (lifestyleType == LifestyleType.Transient)
+			{
+				serviceCollection.AddTransient(serviceType, p =>
+				{
+					container.RequireScope();
+					return container.Resolve(serviceType);
+				});
+			}
+
+			if (lifestyleType == LifestyleType.Scoped)
```
jonorossi: Probably should be an `else if`.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -84,32 +225,40 @@ public static void UseCastleWindsor(this IApplicationBuilder app, IWindsorContai
 			{
 				var resolve = container.Resolve<T>();
 				await resolve.InvokeAsync(context, async (ctx) => await next());
+				container.Release(resolve);
```
jonorossi: It is probably worth a try/finally in case InvokeAsync throws.

src/Castle.Facilities.AspNetCore/Resolvers/FrameworkDependencyResolver.cs
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		public object Resolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
+		{
+			serviceProvider = serviceProvider ?? serviceCollection.BuildServiceProvider();
+			return serviceProvider.GetService(dependency.TargetType);
+		}
+	}
+
+	public static class GenericTypeExtensions
```
jonorossi: Is it a good idea making `GenericTypeExtensions` and `ServiceCollectionExtensions` public? Are you using them in unit tests too?

docs/aspnetcore-facility.md
```diff
@@ -10,38 +10,78 @@ The ASP.NET Core facility provides Castle Windsor integration using a custom act
...
+```csharp
+services.AddSingleton<FrameworkMiddleware>(); // Do this if you don't care about using Windsor
+```
+
+Alternatively if you would like to inject dependencies into your middleware that are registered with Castle Windsor then you can use the 
+`UseCastleWindsorMiddleware` extension.
+
+```csharp
+// Add custom middleware, do this if your middleware uses DI from Windsor
+app.UseCastleWindsorMiddleware<CustomMiddleware>(container);
```
jonorossi: `UseCastleWindsorMiddleware` sounds like it is adding some special Windsor middleware to the pipeline, I was wondering if it should be named something like `UseMiddlewareViaWindsor<TMiddleware>` or `UseMiddlewareFromWindsor<TMiddleware>`?

I know this was there from the first cut but now is the best time to make the change. I was wondering what Autofac named theirs and I can't find how they do this same sort of thing. @fir3pho3nixx you are the expert here so I'll let you decide whether you want to rename it.

docs/aspnetcore-facility.md
```diff
@@ -10,38 +10,78 @@ The ASP.NET Core facility provides Castle Windsor integration using a custom act
...
+```
+
+### Cross Wiring into the IServiceCollection
+
+There is an additional feature you can use to `Cross Wire` components into the IServiceCollection. This is useful for cases where the framework
+needs to know how to resolve a type for example with the Razor @Inject directive in MVC projects. 
+
+```csharp
+container.Register(Component.For<IUserService>().ImplementedBy<AspNetUserService>()
+	.CrossWired(container, services, LifestyleType.Scoped));
```
jonorossi: It sucks that you have to specify the container in `CrossWired` but I know why.

docs/aspnetcore-facility.md
```diff
@@ -10,38 +10,78 @@ The ASP.NET Core facility provides Castle Windsor integration using a custom act
...
-	.BasedOn<Controller>().LifestyleScoped()); // <- Change lifestyle
-container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly)
-	.BasedOn<ViewComponent>().LifestyleTransient());
-container.Register(Classes.FromAssemblyInThisApplication(typeof(TStartup).Assembly)
-	.BasedOn<TagHelper>().LifestyleTransient());
+services.AddCastleWindsor(container, opts =>
+{
+	opts.RegisterControllers(typeof(HomeController).Assembly, LifestyleType.Transient);
+	opts.RegisterTagHelpers(typeof(EmailTagHelper).Assembly, LifestyleType.Transient);
+	opts.RegisterViewComponents(typeof(AddressComponent).Assembly, LifestyleType.Transient);
```
jonorossi: I'm sure you thought about it or used a pattern from somewhere, but could you explain why the class that you perform actions on is named "options".

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -29,44 +29,185 @@ namespace Castle.Facilities.AspNetCore
...
+				catch
+				{
+					entryAssembly = entryAssembly ?? Assembly.GetCallingAssembly();
+				}
+
+				return entryAssembly;
+			}
+		}
+
+		internal bool ControllersAlreadyRegistered => controllersAlreadyRegistered;
```
jonorossi: Why not just define internal auto properties with a private setter?

docs/aspnetcore-facility.md
fir3pho3nixx: This should be private. Agreed.

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
fir3pho3nixx: Good spot, will remove.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: I will do. It is just lambda parameter based null coalescence to avoid null reference exceptions. Is this bad? Can I still use null coalescence for invalid arguments?

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: Yes we should design by contract. Do you know of any way we could integrate the other types? I personally have never needed them in a web scenario.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: Or do we throw are LifestyleNotSupportedException at https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs#L186?

We need to figure out https://github.com/castleproject/Windsor/pull/394#discussion_r183031893 before we decide this.

src/Castle.Facilities.AspNetCore/Resolvers/FrameworkDependencyResolver.cs
fir3pho3nixx: No, none at all. Purely muscle memory kicking in here.

docs/aspnetcore-facility.md
fir3pho3nixx: @jonorossi I really like this. It makes it discoverable via code completion. 

I love `UseMiddlewareViaWindsor<TMiddleware>`, the responsible choice is definitely `UseMiddlewareFromWindsor<TMiddleware>`.

I am happy with how we are refining the UX here.

docs/aspnetcore-facility.md
fir3pho3nixx: Yup, this sucked hard.

fir3pho3nixx: Any suggestions?

docs/aspnetcore-facility.md
fir3pho3nixx: Yeah, this is what it means from an HTTPS point of view.

```
public static void Main(string[] args)
{
	BuildWebHost(args).Run();
}

public static IWebHost BuildWebHost(string[] args) =>
WebHost.CreateDefaultBuilder(args)
	.UseStartup<Startup>()
	.UseKestrel(options =>
	{
		options.Listen(IPAddress.Loopback, 5000);
		options.Listen(IPAddress.Loopback, 5001, listenOptions =>
	{
		listenOptions.UseHttps("testCert.pfx", "testPassword");
	});
	}).Build();
```

fir3pho3nixx: It is a convention.

fir3pho3nixx: We can clean this up to make options more stateful but it would mean adding more code.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: Yep can do.

src/Castle.Facilities.AspNetCore/Resolvers/FrameworkDependencyResolver.cs
fir3pho3nixx: These are super covered by tests in https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore.Tests/AspNetCoreFacilityTestCase.cs

fir3pho3nixx: And https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore.Tests/Resolvers/ServiceCollectionExtensionsTestCase.cs#L50

fir3pho3nixx: > Are you using them in unit tests too?
No, I will move this over ...

fir3pho3nixx: It is integration tested by:
 - https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore.Tests/AspNetCoreFacilityTestCase.cs
 - https://github.com/fir3pho3nixx/Windsor/blob/aspnet-core-windsor-final/src/Castle.Facilities.AspNetCore.Tests/Resolvers/ServiceCollectionExtensionsTestCase.cs

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
jonorossi: > Can I use null coalescence for invalid arguments?

Yes, definitely, those look great to handle the throw exception case.

> It is just lambda parameter based null coalescence to avoid null reference exceptions. Is this bad?

Other than readability my concern is that doing it this way actually runs more code at runtime and that empty lambda gets emitted as a compiler named method.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
jonorossi: > Do you know of any way we could integrate the other types?

Someone could create any sort of custom lifestyle type, not sure you'll be able to handle that with this facility, probably just worth throwing on unsupported lifestyle until we need them and have a proper use case.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
jonorossi: Yep, throw that exception in the else statement 👍.

src/Castle.Facilities.AspNetCore/Resolvers/FrameworkDependencyResolver.cs
jonorossi: > No, I will move this over ...

No problem. I wasn't worried about them being unit tested, more wondering if they were public because the unit tests also used them.

docs/aspnetcore-facility.md
jonorossi: Great, especially that it'll be more discoverable.

Were you planning to name it `UseMiddlewareFromWindsor` or `UseMiddlewareFromCastleWindsor`. I was wondering if the "Castle" bit should be dropped from `AddCastleWindsor` and `UseCastleWindsor` since it doesn't really add anything useful?

docs/aspnetcore-facility.md
jonorossi: I'll leave it up to you to choose but it just felt a little strange that it was actually doing something rather than just holding info on what the user wanted until calling `AddCastleWindsor` just like like the above does start listening until calling `UseKestrel` or actually probably in `WebHost.Run`.

Couldn't you change `AddApplicationComponentsToWindsor` so rather than looking at boolean members like `ControllersAlreadyRegistered`, have members typed `List<Pair<Assembly, LifeStyleType>>`? It should be about the same amount of code and that way it has the register code once inside `AddApplicationComponentsToWindsor` and it'll either do the stored list or use `options.EntryAssembly`.

docs/aspnetcore-facility.md
jonorossi: Other than being a little ugly, one problem with this approach is that `CrossWired` registers the service in the ASP.NET Core container before it is even registered in Windsor.

What about using a ComponentModel construction contributor (`IContributeComponentModelConstruction`) that does the cross wiring after the `ComponentModel` is built. The WCF facility has one (`WcfBehaviorInspector`) to set up a custom activator, as do most facilities including typed factory, startable, event wiring, synchronize, and built in things like mixins.

docs/aspnetcore-facility.md
fir3pho3nixx: Done

src/Castle.Facilities.AspNetCore/Castle.Facilities.AspNetCore.csproj
fir3pho3nixx: Done

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: Done

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: Done

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: Yep. Good shout. Done.

docs/aspnetcore-facility.md
fir3pho3nixx: You will be pleased to know, `UseCastleWindsor` is not a thing anymore. We now only need `UseWindsor` :). Done.

fir3pho3nixx: Oh and `UseMiddlewareFromWindsor` is the new name for the middleware extension.

docs/aspnetcore-facility.md
fir3pho3nixx: > Other than being a little ugly, one problem with this approach is that CrossWired registers the service in the ASP.NET Core container before it is even registered in Windsor.

This raises a really important question about how people use Windsor. I hung this extension off ComponentRegistration<T> because I assumed it would be called within the context of Container.Register(where I might be getting it wrong). The cross wiring here is merely registering late bound components with factories in the IServiceCollection. My assumption was eventually the service will be registered in both places. 

> What about using a ComponentModel construction contributor 

Yes, this would be hard. What CrossWiring is really trying to solve, is making types registered through Windsor available to framework code buried behind IServiceCollection/ServiceProvider logic like the MVC Razor @Inject directive. This is a registration time consideration.  I am worried that construction contribution does not consider the registration requirements and anything you can do during resolve to implement this might already be too late.

That said, I still think we should wrangle with this one a little more to tidy it up. I am just not sure what that would be yet. 

If you have any more questions about cross wiring, please fire away.

docs/aspnetcore-facility.md
fir3pho3nixx: Do you know what, this was not that bad. I added it in. It did bloat the `AddApplicationComponentsToWindsor` a little but not as bad as I originally thought.

fir3pho3nixx: Done

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: These are gone now, but yes, the new implementation works this way.

docs/aspnetcore-facility.md
fir3pho3nixx: > Other than being a little ugly

Yes, it is nasty.

src/Castle.Facilities.AspNetCore/Exceptions/LifestyleNotSupportedException.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Castle.Facilities.AspNetCore.Exceptions
+{
+	using System;
+
+	using Castle.Core;
+
+	/// <summary>
+	/// For when an unsupported <see cref="LifestyleType"/> get passed to the <see cref="CastleWindsorRegistrationExtensions.CrossWired{TService}(Castle.MicroKernel.Registration.ComponentRegistration{TService},Castle.Windsor.IWindsorContainer,Microsoft.Extensions.DependencyInjection.IServiceCollection,Castle.Core.LifestyleType)"/> extension method.
+	/// </summary>
+	public class LifestyleNotSupportedException : Exception
```
jonorossi: Is a custom exception really necessary for a situation no one is going to specifically catch, I thought you were just going to use a `NotSupportedException` or even better a `ArgumentException`.

docs/aspnetcore-facility.md
jonorossi: > Yes, this would be hard. What CrossWiring is really trying to solve, is making types registered through Windsor available to framework code buried behind IServiceCollection/ServiceProvider logic like the MVC Razor @Inject directive. This is a registration time consideration. I am worried that construction contribution does not consider the registration requirements and anything you can do during resolve to implement this might already be too late.

Not sure I follow why this would be hard, I understand that you are doing late bound components but ComponentModel construction contributors are run directly after construction of the ComponentModel (before `container.Register()` returns), not at resolve time, see [the docs](https://github.com/castleproject/Windsor/blob/master/docs/componentmodel-construction-contributors.md).

The WCF facility uses `ComponentRegistration` to set up services, and after they are registered in the container it'll do a bunch of other stuff. The Startable facility is just the same, you don't have to implement `IStartable` and can use [the `ComponentRegistration` methods](https://github.com/castleproject/Windsor/blob/master/docs/startable-facility.md) and its CMCC adds a start/stop concern if a new component is startable.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
jonorossi: `configure?.Invoke(options)` is much better, and it happens that it is the exact same IL with the other syntax.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -29,44 +32,172 @@ namespace Castle.Facilities.AspNetCore
 	using Microsoft.AspNetCore.Razor.TagHelpers;
 	using Microsoft.Extensions.DependencyInjection;
 
+	/// <summary>
+	/// For overriding default registration and lifestyles behaviour
+	/// </summary>
+	public class CastleWindsorRegistrationOptions
```
jonorossi: Should these classes also drop the "Castle" bit?

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
```diff
@@ -77,39 +208,68 @@ public static void UseCastleWindsor(this IApplicationBuilder app, IWindsorContai
...
+			}
+
+			if (!options.TagHelperRegistrations.Any())
+			{
+				container.Register(Classes.FromAssemblyInThisApplication(options.EntryAssembly).BasedOn<ViewComponent>().LifestyleScoped());
+			}
+
+			foreach (var controllerRegistration in options.TagHelperRegistrations)
+			{
+				container.Register(Classes.FromAssemblyInThisApplication(controllerRegistration.Item1).BasedOn<Controller>().Configure(x => x.LifeStyle.Is(controllerRegistration.Item2)));
```
jonorossi: Should this and view components be based on a different class other than Controller, I think you had this right before you recent changes.

docs/aspnetcore-facility.md
fir3pho3nixx: Ok, let’s try it this way round. What would they contribute to resolves from IServiceCollection/IServiceProvider?

fir3pho3nixx: I have seen this code and will definitely give it another look. I am just trying to figure how Windsor abstractions will help here. The problem is on the MS side. Please bear with me.

src/Castle.Facilities.AspNetCore/Exceptions/LifestyleNotSupportedException.cs
fir3pho3nixx: Thought a custom exception with a namespace and a custom message would make the stack trace a little cleaner. Will change it back.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: Yup. Definitely.

src/Castle.Facilities.AspNetCore/CastleWindsorRegistrationExtensions.cs
fir3pho3nixx: And here come some tests. The history of this code is something that vexed me.

fir3pho3nixx: Really vexed me.

generik0:
Great work guys. This is going to be really really Great!

Any thoughts of the Best way to get the model binders to work with the IOC? 
[custom model(https://docs.microsoft.com/en-us/aspnet/core/mvc/advanced/custom-model-binding?view=aspnetcore-2.1)
I am thinking especially about this, ie a binding provider.

Would it be best to just container.Resolve it?

```c#
public void ConfigureServices(IServiceCollection services)
{
    services.AddDbContext<AppDbContext>(options => options.UseInMemoryDatabase());

    services.AddMvc(options =>
    {
        // add custom binder to beginning of collection
        options.ModelBinderProviders.Insert(0, new AuthorEntityBinderProvider());
    });
}
```

generik0:
Oh yes, and in here use a factory to resolve the binder?

Or do you think I could make my own resolver instead to the provider?

```c#
public class AuthorEntityBinderProvider : IModelBinderProvider
    {
        public IModelBinder GetBinder(ModelBinderProviderContext context)
        {
            if (context == null)
            {
                throw new ArgumentNullException(nameof(context));
            }

            if (context.Metadata.ModelType == typeof(Author))
            {
                return new BinderTypeModelBinder(typeof(AuthorEntityBinder));
            }

            return null;
        }
    }
```

fir3pho3nixx:
@generik0, let me check and get back to you. My suspicion is that this does not get run through the ServiceProvider so CrossWiring might not be an option. The same problem would exist for filters.

fir3pho3nixx:
@generik0 I have a question. I have typically used model binding for converting inputs to controllers. Why do you need an IoC abstraction over this? I am just incredibly curious to know why this is even happening on your end. I have used SOA patterns for enrichment but we usually push this out to the backend. What is happening in your view that you need this abstraction?

generik0:
@fir3pho3nixx thanks for the questions. It isn’t actually for a view. In aspnetcore (as you probably know) there is no longer a api controller, it is all just controller.
If you want to send more than one data object in a form body from the frontend (a Spa frontend) through a post call you need to do the bindings (as far as I could read).
In asp.net I used NancyFx bindings for the Api, and it was no problem doing these kind of bindings.
Aspnetcore it seams plausible also, just inst integrated (by MS) with the DI :-/

generik0:
@fir3pho3nixx should I remove these questions from the PR here and make a separate issue?

generik0:
@fir3pho3nixx  update about providers. Found this link, hence i think the answer is simple and as suspected. I need to do a cross wire of sorts and let then container resolve the provider.
In the provider, use factories for anything that needs to be resolver transient or scope


https://stackoverflow.com/questions/44598683/how-to-use-simple-injector-to-inject-a-class-into-my-custom-model-binder-provide?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

fir3pho3nixx:
> @fir3pho3nixx update about providers. Found this link, hence i think the answer is simple and as suspected. I need to do a cross wire of sorts and let then container resolve the provider.
In the provider, use factories for anything that needs to be resolver transient or scope

Looks like the answer is to use service location. In our implementation you should not need any cross wiring, register it in a vanilla way with Windsor. Should work. I would expect the same thing would need to happen for filters.

@generik0 

After the chat on issue #120, I did manage to fix a whole heap of problems with our implementation.

 - Using the BuilderServiceProvider method in the resolvers was in fact new'ing up a new container. This was definitely causing singletons to double up because aspnet/hosting was already creating one. After debugging the Startup.ConfigureServices method all the way down to hosting, I noticed the safest way forward here was to return it from the AddWindsor extension with an optional factory method where the user can create it themselves if they choose. 

 - Disposables for cross wired component's were all over the place. In some cases things were getting disposed twice and in other cases not at all. This must have caused a few headaches on your end. I added a significant amount of unit testing around this to fix it. 

 - We still have to disable scope validation for the ServiceProvider, unfortunate but if we are to do this properly then we would have to start managing 2 sets of scopes. I am hesitant to add this in just yet until I can research the problem slightly better. 

@jonorossi 

I think we have bottomed out the torn lifestyle problem. There are some quirky looking unit test's in here but they validate my observations from running it in https://github.com/fir3pho3nixx/Windsor.AspNetCore. There ain't no free lunches up in here that is for sure :) I think I am ready to start finishing the work of improving the cross wire extension to make it a lot less sucky. Should get to it next week.

docs/aspnetcore-facility.md
```diff
@@ -8,40 +8,80 @@ The ASP.NET Core facility provides Castle Windsor integration using a custom act
...
+when web requests are made to the server for TagHelpers, ViewComponents and Controllers. 
 
-This method also adds a sub resolver for dealing with the resolution of ASP.NET Core framework 
-types. An example might be something like an [ILoggerFactory](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.iloggerfactory?view=aspnetcore-2.0).
-It is important to note that this extension also injects custom middleware for the management of scoped lifestyles. This middleware calls
-the `WindsorContainer.BeginScope` extension. It will also dispose the scope once the request is completed.
+This method also adds a sub resolver for dealing with the resolution of ASP.NET Core framework types. An example might be something like an 
+[ILoggerFactory](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.iloggerfactory?view=aspnetcore-2.0). 
+It is important to note that this extension also injects custom middleware for the management of implicitly scoped lifestyles. It will also 
+dispose the implicit scope once the request is completed.
```
fir3pho3nixx: Might need to clarify that these are windsor scopes.

src/Castle.Facilities.AspNetCore.Tests/Castle.Facilities.AspNetCore.Tests.csproj
```diff
@@ -20,4 +20,23 @@
 	  <ProjectReference Include="..\Castle.Facilities.AspNetCore\Castle.Facilities.AspNetCore.csproj" />
 	</ItemGroup>
 
+	<ItemGroup>
+	  <None Update="Fakes\ModelFakes.tt">
+	    <Generator>TextTemplatingFileGenerator</Generator>
```
fir3pho3nixx: @jonorossi hope you are OK with this, otherwise I will convert this to code files.

src/Castle.Facilities.AspNetCore.Tests/Framework/Builders/ServiceCollectionBuilder.cs
```diff
@@ -0,0 +1,30 @@
...
+	using Castle.Facilities.AspNetCore.Tests.Fakes;
+
+	using Microsoft.Extensions.DependencyInjection;
+
+	public class ServiceCollectionBuilder
+	{
+		public static ServiceCollection New()
+		{
+			var serviceCollection = new ServiceCollection();
+			ModelInstaller.RegisterServiceCollection(serviceCollection);
```
fir3pho3nixx: Use of code generation

src/Castle.Facilities.AspNetCore.Tests/Framework/Builders/ServiceProviderBuilder.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+	using System;
+
+	using Microsoft.Extensions.DependencyInjection;
+
+	public class ServiceProviderBuilder
+	{
+		public static IServiceProvider New(IServiceCollection services)
+		{
+			return services.BuildServiceProvider(validateScopes: false);
```
fir3pho3nixx: I would like to figure out what types are causing this to fail in the runtime tests @ https://github.com/fir3pho3nixx/Windsor.AspNetCore.

src/Castle.Facilities.AspNetCore.Tests/Framework/Builders/WindsorContainerBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+			var container = new WindsorContainer();
+			RegisterApplicationComponents(container, services);
+			services.AddWindsor(container, configure, serviceProviderFactory);
+			return container;
+		}
+
+		private static void RegisterApplicationComponents(WindsorContainer container, IServiceCollection serviceCollection)
+		{
+			ModelInstaller.RegisterWindsor(container);
+			ModelInstaller.RegisterCrossWired(container, serviceCollection);
```
fir3pho3nixx: Use of code generation

src/Castle.Facilities.AspNetCore/Resolvers/FrameworkDependencyResolver.cs
```diff
@@ -0,0 +1,85 @@
...
+
+		public FrameworkDependencyResolver(IServiceCollection serviceCollection)
+		{
+			this.serviceCollection = serviceCollection;
+		}
+
+		public void AcceptServiceProvider(IServiceProvider serviceProvider)
+		{
+			this.serviceProvider = serviceProvider;
+		}
```
fir3pho3nixx: Required to not create torn lifestyles.

src/Castle.Facilities.AspNetCore/Resolvers/IAcceptServiceProvider.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Castle.Facilities.AspNetCore.Resolvers
+{
+	using System;
+
+	public interface IAcceptServiceProvider
+	{
+		void AcceptServiceProvider(IServiceProvider serviceProvider);
+	}
+}
```
fir3pho3nixx: Required to not create torn lifestyles.

src/Castle.Facilities.AspNetCore/Resolvers/LoggerDependencyResolver.cs
```diff
@@ -14,21 +14,22 @@
...
-		private readonly ILoggerFactory loggingFactory;
+		private IServiceProvider serviceProvider;
 
-		public LoggerDependencyResolver(IServiceCollection serviceCollection)
+		public void AcceptServiceProvider(IServiceProvider serviceProvider)
 		{
-			var serviceProvider = serviceCollection.BuildServiceProvider();
-			this.loggingFactory = serviceProvider.GetService<ILoggerFactory>();
+			this.serviceProvider = serviceProvider;
 		}
```
fir3pho3nixx: Required to not create torn lifestyles.

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,213 @@
...
+			if (lifestyleType == LifestyleType.Transient)
+			{
+				serviceCollection.AddTransient(serviceType, p =>
+				{
+					return container.Resolve(serviceType);
+				});
+			}
+			else if (lifestyleType == LifestyleType.Scoped)
+			{
+				registration.Activator<CrossWiredComponentActivator>();
```
fir3pho3nixx: Scoped cross wired components are released by implicit scopes within the ServiceProvider not the WindsorContainer.

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,213 @@
...
+				serviceCollection.AddScoped(serviceType, p =>
+				{
+					return container.Resolve(serviceType);
+				});
+			}
+			else if (lifestyleType == LifestyleType.Singleton)
+			{
+				registration.Activator<CrossWiredComponentActivator>();
```
fir3pho3nixx: Singleton cross wired components are released by disposing the ServiceProvider not the WindsorContainer.

src/Castle.Facilities.AspNetCore/Resolvers/LoggerDependencyResolver.cs
```diff
@@ -14,21 +14,22 @@
...
+	public class LoggerDependencyResolver : ISubDependencyResolver, IAcceptServiceProvider
 	{
-		private readonly ILoggerFactory loggingFactory;
+		private IServiceProvider serviceProvider;
 
-		public LoggerDependencyResolver(IServiceCollection serviceCollection)
+		public void AcceptServiceProvider(IServiceProvider serviceProvider)
 		{
-			var serviceProvider = serviceCollection.BuildServiceProvider();
-			this.loggingFactory = serviceProvider.GetService<ILoggerFactory>();
```
fir3pho3nixx: Required to not create torn lifestyles.

fir3pho3nixx:
We have jumped from 8 unit tests to [137](https://ci.appveyor.com/project/castleproject/windsor/build/0.0.417#L359).

src/Castle.Facilities.AspNetCore.Tests/Castle.Facilities.AspNetCore.Tests.csproj
fir3pho3nixx: I wish I could replace some of the fluff in Windsor with this. Man there is a whole heap of test boiler plate hanging around.

src/Castle.Facilities.AspNetCore.Tests/Framework/Builders/ServiceProviderBuilder.cs
fir3pho3nixx: Quick check revealed transients,

fir3pho3nixx:
@jonorossi - I have fixed the cross wiring API to be much cleaner. The component model contributor with custom activator turned out to hold the answer all along. Thanks. 

Could you please take another look?

fir3pho3nixx:
@generik0 are you happy with this cut? We are dog fooding and are running this in prod. @jonorossi has a lot of traffic on his hands. I don’t expect he will humanly be able to get to this ... won’t merge unless you say it is ok.

generik0:
Hi @fir3pho3nixx.
I have been killing myself the last three weeks with a major release @work.
But i beleive I will have time to look at it in the weekend  :-)

fir3pho3nixx:
@generik0 thanks for having a look, I think we are nearly there. I am going to start squishing these commit's down as soon as I get some free time.

generik0:
No thank you @fir3pho3nixx 
Really great work. Looks nice and clean. As clean as can be taking into considerations the limitations posed by aspnetcore / MS DI. Really great work. 
I am integrating the framework as we speak. I will report back if i find any isses.

Thanks again, good work! Regards, G.

fir3pho3nixx:
@jonorossi 

Could you please kindly take a look. Think we are done here for now.

src/Castle.Facilities.AspNetCore/Contributors/CrossWiringComponentModelContributor.cs
```diff
@@ -0,0 +1,61 @@
...
+		private readonly IServiceCollection services;
+
+		public CrossWiringComponentModelContributor(IServiceCollection services)
+		{
+			this.services = services;
+		}
+
+		public void ProcessModel(IKernel kernel, ComponentModel model)
+		{
+			if (model.Configuration.Attributes.Get("aspnet-core-cross-wired") == "yes")
```
jonorossi: I know some of the facilities don't but it would be good to use a constant for `aspnet-core-cross-wired`, consider even making a [public readonly field like the typed factory](https://github.com/castleproject/Windsor/blob/bbc1183a8806f4cfdf07dc462e1895313fb0684a/src/Castle.Windsor/Facilities/TypedFactory/TypedFactoryFacility.cs#L39-L48).

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,184 @@
...
+		}
+
+		/// <summary>
+		/// For making services registered in Windsor visible to the IServiceCollection using 'late bound' factory methods. This makes things like the @Inject directive in Razor work.
+		/// </summary>
+		/// <param name="registration">The component registration that gets copied across to the IServiceCollection</param>
+		public static ComponentRegistration CrossWired(this ComponentRegistration registration)
+		{
+			registration.Activator<CrossWiredComponentActivator>();
+			registration.Attribute("aspnet-core-cross-wired").Eq("yes");
```
jonorossi: Just in case you didn't already know, there is an `AddAttributeDescriptor` method that does the same thing. You could also use `bool.TrueString` as your constant if you like.

```
componentRegistration
    .AddAttributeDescriptor(TypedFactoryFacility.IsFactoryKey, bool.TrueString);
```

src/Castle.Facilities.AspNetCore/Activators/CrossWiredComponentActivator.cs
```diff
@@ -0,0 +1,47 @@
...
+		/// </summary>
+		/// <param name="steps">All decomission concerns are fine except DisposalConcern</param>
+		/// <param name="instance">The cross wired instance</param>
+		protected override void ApplyConcerns(IEnumerable<IDecommissionConcern> steps, object instance)
+		{
+			if (Model.LifestyleType == LifestyleType.Scoped || Model.LifestyleType == LifestyleType.Singleton)
+			{
+				foreach (var concern in steps.Where(x => x.GetType() != typeof(DisposalConcern)))
+				{
+					concern.Apply(Model, instance);
```
jonorossi: I assume you weren't able to just remove the `DisposalConcern` from `ComponentModel.Lifecycle` in `CrossWiringComponentModelContributor`?

Will `LateBoundDecommissionConcerns` also cause the same problem? I guess those can't really be used with this facility as it needs to know the component type upfront to cross wire it.

fir3pho3nixx:
> @fir3pho3nixx just a few more comments 😄.

I know, I am simultaneously being pulled up by some good guys all over the place. I really would not have it any other way. Thanks.

src/Castle.Facilities.AspNetCore/Contributors/CrossWiringComponentModelContributor.cs
fir3pho3nixx: Hint accepted. I promised a first class API(excluding typed factories and parent child containers). LOL.

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,184 @@
...
+		/// <summary>
+		/// This exposes Windsor components to the IServiceCollection using the `CrossWired` registration extension. 
+		/// </summary>
+		/// <param name="services">ASP.NET Core service collection from Microsoft.Extensions.DependencyInjection</param>
+		/// <param name="container">Windsor container which IServiceCollection resolves from using factory methods</param>
+		public static void CrossWiresInto(this IWindsorContainer container, IServiceCollection services)
+		{
+			InstallCrossWiringComponentModelContributor(container.Kernel, services);
+		}
+
```
fir3pho3nixx: @generik0 also thought we needed to throw a helpful exception here in case somebody used the [CrossWire](https://github.com/castleproject/Windsor/pull/394/files#diff-77a22f5542ffd29a9b211c43bfb976f3R83) extension.

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,184 @@
...
+	public static class WindsorRegistrationExtensions
+	{
+		/// <summary>
+		/// This exposes Windsor components to the IServiceCollection using the `CrossWired` registration extension. 
+		/// </summary>
+		/// <param name="services">ASP.NET Core service collection from Microsoft.Extensions.DependencyInjection</param>
+		/// <param name="container">Windsor container which IServiceCollection resolves from using factory methods</param>
+		public static void CrossWiresInto(this IWindsorContainer container, IServiceCollection services)
+		{
```
fir3pho3nixx: @generik0 also thought we needed to throw a helpful exception here in case somebody used the [CrossWire](https://github.com/castleproject/Windsor/pull/394/files#diff-77a22f5542ffd29a9b211c43bfb976f3R83) extension without calling this.

src/Castle.Facilities.AspNetCore/Contributors/CrossWiringComponentModelContributor.cs
fir3pho3nixx: ( Now I know enough to do this really easy :) )

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
fir3pho3nixx: I agree.

src/Castle.Facilities.AspNetCore/Contributors/CrossWiringComponentModelContributor.cs
jonorossi: There is no need for an `IFacility` implementing class, just find a sensible place for the constant, probably `WindsorRegistrationExtensions`.

src/Castle.Facilities.AspNetCore/Activators/CrossWiredComponentActivator.cs
fir3pho3nixx: We are cool with perf testing here: https://github.com/fir3pho3nixx/Windsor.AspNetCore/commit/815221ed20cd9388fc7f89a38a04b4642163c511

fir3pho3nixx: Fuck sake. We have a memory leak. 

Just watched it grow from 220mb to 350mb using typed factories. Che cristo.

fir3pho3nixx: @jonorossi it is not fair I eat this on this PR. I have been quite vocal about typed factories. They are rubbish for many reasons.  Is it fair to say I do not support typed factories in this PR? Can I sort that next? Not cool that I get held to account for ex-emeritus.

fir3pho3nixx: Please see: https://github.com/castleproject/Windsor/issues/410

My confidence is knocked.

jonorossi: > We have a memory leak.
>
> it is not fair I eat this on this PR. I have been quite vocal about typed factories. They are rubbish for many reasons. Is it fair to say I do not support typed factories in this PR? Can I sort that next? Not cool that I get held to account for ex-emeritus.
>
> ** Where are you guys anyway, help now right or I will re-write your shit **

I'm not sure exactly why you are getting so aggressive here, I get that this pull request has gone on a while. Just a polite reminder the only active committers are those listed on the team page, there should be no expectation of others contributing.

That aside, we can definitely merge this not supporting typed factories, we can even release this without supporting typed factories if you want, remember we get to choose the direction and releasing software with known defects or limitations isn't the end of the world.

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
```diff
@@ -0,0 +1,198 @@
...
+
+	using Microsoft.AspNetCore.Builder;
+	using Microsoft.AspNetCore.Http;
+	using Microsoft.AspNetCore.Mvc;
+	using Microsoft.AspNetCore.Razor.TagHelpers;
+	using Microsoft.Extensions.DependencyInjection;
+
+	public static class WindsorRegistrationExtensions
+	{
+		private static bool crossWiresIntoWasCalledBeforeRegistration = false;
```
jonorossi: Doesn't this have to be called for each container, so this being static wouldn't work for a second container in the process?

src/Castle.Facilities.AspNetCore/Activators/CrossWiredComponentActivator.cs
fir3pho3nixx: Thanks @jonorossi. I still believe we need to support typed factories but I know from our issue tracker that comes with even longer tail. 

As long as we know what the problems are before we release I am good. 

Aside, we had our first use case for a cross wire today and I dropped this PR in and it just worked! Very happy with the current state of this.

src/Castle.Facilities.AspNetCore/WindsorRegistrationExtensions.cs
fir3pho3nixx: Awesome, I need a way of tacking a reference to the IServiceCollection on the container. I was hunting through the internals of the IKernel but then I just got grumpy and tired. 

Is there anyway this can be done? That way we don’t smash container life cycles.

Also, there can be only 1 service provider when it comes to the hosting runtime, but I don’t see why we can’t have multiple Windsor containers. 

You OK with me adding a general purpose key/value store to the kernel to solve this?

jonorossi: > You OK with me adding a general purpose key/value store to the kernel to solve this?

Not really. This is why most facilities will have a facility instance which can hold whatever extra variables it needs and set up the facility in the container, I don't think we want to pollute the container with a generic KV store when we've got a pattern that works well.

I guess if you didn't want that you could have another component model contributor for this facility to check the configuration when you register, and check `kernel.ComponentModelBuilder.Contributors.Any(c => c.GetType() == typeof(CrossWiringComponentModelContributor))`, however you'd then be better off just always installing the `CrossWiringComponentModelContributor` and allowing `services` to be null, and `CrossWiringComponentModelContributor` will throw when you call `Register` not when you call `CrossWired` since it doesn't yet know which container the `ComponentRegistration` is going to be registered in. Even with a KV store on the container you don't know the container until `Register` is called, so this has to change.

fir3pho3nixx: Ok in this case we should try housing the ServiceCollection reference in a facility first. I am almost 100% certain this will work. 

We can use the contributor as a fallback position if we don’t like the way it looks. 

Thanks for explaining.

fir3pho3nixx: @generik0 

I will submit a review. Can you give it one last look. I will highlight the latest changes.

docs/aspnetcore-facility.md
```diff
@@ -8,40 +8,92 @@ The ASP.NET Core facility provides Castle Windsor integration using a custom act
...
+```
+
+For this to work you have add the facility in the configure services method before you register anything. A helpful exception
+will be thrown here, in case you forget.
+
+```csharp
+public IServiceProvider ConfigureServices(IServiceCollection services)
+{
+	// Setup component model contributors for making windsor services available to IServiceProvider
+	Container.AddFacility<AspNetCoreFacility>(f => f.CrossWiresInto(services));
```
fir3pho3nixx: @generik0 

Important breaking change. This is what eventually unlocked the key of being able to treat the WindsorContainer and the IServiceCollection as a tuple.

src/Castle.Facilities.AspNetCore.Tests/WindsorRegistrationExtensionsAcceptanceTestCase.cs
```diff
@@ -0,0 +1,37 @@
...
+	[TestFixture, Order(0)]
+	public class WindsorRegistrationExtensionsAcceptanceTestCase
+	{
+		[Test, Order(0)]
+		public void Should_throw_for_CrossWire_Registration_without_CrossWiresInto_being_called_first()
+		{
+			using (var container = new WindsorContainer())
+			{
+				Assert.Throws<InvalidOperationException>(() => { container.Register(Component.For<AnyComponent>().CrossWired()); });
+			}
```
fir3pho3nixx: @generik0 

Your suggestion of a helpful exception was implemented here. Thanks!

src/Castle.Facilities.AspNetCore/Activators/CrossWiredComponentActivator.cs
fir3pho3nixx: The memory leak is small, but it is there. I am updated the docs to reflect this.

src/Castle.Facilities.AspNetCore/AspNetCoreFacility.cs
```diff
@@ -0,0 +1,48 @@
...
+	using Castle.Facilities.AspNetCore.Contributors;
+	using Castle.MicroKernel.Facilities;
+
+	using Microsoft.Extensions.DependencyInjection;
+
+	public class AspNetCoreFacility : AbstractFacility
+	{
+		internal const string IsCrossWiredIntoServiceCollectionKey = "windsor-registration-is-also-registered-in-service-collection";
+
+		public static readonly AsyncLocal<AspNetCoreFacility> Current = new AsyncLocal<AspNetCoreFacility>();
```
fir3pho3nixx: @jonorossi 

Hope you don't mind the ambient state here. Most of the registration, happens in a single thread, so I think we are good. I am a bit worried about doing this as I wont be `dog fooding` this on net45.

docs/aspnetcore-facility.md
```diff
@@ -127,30 +208,39 @@ public sealed class CustomMiddleware : IMiddleware
...
+### Typed Factories
+
+Typed factories are supported although there are some problems we do know about. 
+
+```csharp
+Container.Register(Component.For<LateBoundTransientDisposable>().UsingFactoryMethod(() => new LateBoundTransientDisposable()).LifestyleTransient().CrossWired());
+```
+
+After rigourous testing we noticed a small memory leak, we are into typed factories here: https://github.com/castleproject/Windsor/issues/410
+
```
fir3pho3nixx: @jonorossi && @generik0 

Small memory leak here. Just thought we would call that out. The process creeps up at a glacial pace. We are talking Kb's every hour or so(framework??). Either way, if people are picking this up and using typed factories they should at least have a warning of the stuff we know about.

fir3pho3nixx:
@jonorossi && @generik0 

I am not touching this merge button until you guys have had a look.

@anyonewatching

Sorry I am delaying this, but this needs to be right(or as right as it can be) before we release.

generik0:
@fir3pho3nixx yes I am watch/following all the good work.
I will add the changes tonight and report back.
You have written there is a breaking change. Do I need to do any special changes in my registration?
I will look at you test project for guidens if you don’t have time to answer :-)

generik0:
@fir3pho3nixx if there is a leak with type factories, should we be using it?
Is it because of the tracking and objects not getting disposed?
Seriously, if there is a memory leak, I think the previous solution was better :-/

fir3pho3nixx:
> Do I need to do any special changes in my registration?

Very small change to one line of code on startup. 

```csharp
Container.AddFacility<AspNetCoreFacility>(f => f.CrossWiresInto(services));
```

The rest should all be the same. 

> @fir3pho3nixx if there is a leak with type factories, should we be using it? Is it because of the tracking and objects not getting disposed? 

It is small but I do notice it creeping up a few Kb's every hour or so. I need a longer running test to be sure. I will be diving into typed factories next to understand this better. 

> Seriously, if there is a memory leak, I think the previous solution was better :-/

It was always there, we just never tested in anger with late bound component registrations(or typed factories).

src/Castle.Facilities.AspNetCore/AspNetCoreFacility.cs
fir3pho3nixx: > I don't think it makes sense to add the AspNetCoreFacility if and only if you are doing cross-wiring, otherwise it throws if you aren't doing cross-wiring.

I hear you. We can definitely wind this back to how it was. No problem. Before we do I just want to make sure I understand this better.

> I still think this static stuff is completely unnecessary and that the ComponentRegistration should be as stateless as possible, and that the "cross wiring not configured" exception should be throw by CrossWiringComponentModelContributor.ProcessModel after the user calls Register on a container instance, at this point the CMC has access to the IKernel and easy access to AspNetCoreFacility.Services of the facility instance.

For cross wiring, the CMC needs to be installed in the first place and secondly requires a reference to the IServiceCollection. If the CMC was not installed in the first place how would throwing the helpful exception from inside it help? 

Thanks for the tip on the ProcessModel. This would clean up the static reference, I still cannot see how it guard a CrossWire registration from happening unless the CMC was installed in the first place. Does this make sense?

jonorossi: > We can definitely wind this back to how it was. No problem. Before we do I just want to make sure I understand this better.

I actually meant to take it further, users always add the facility, and the facility always adds the CWCMC and when it comes across a component that is set for cross wiring it sets it up, or throws if the container isn't configured correctly, the exception will wind up the stack through container.Register.

fir3pho3nixx: > I actually meant to take it further

Awesome. I like where this is going. 

The original CrossWire extension could either extend the IServiceCollection and take the IWindsorContainer as a parameter or vice versa. This was awful. Introducing the CWCMC really went far to solve this problem. It did move the validation problem somewhere else though.

> users always add the facility

What we are validating against here is whether someone added the facility or not. Now we have a facility with a configuration callback. 

> and the facility always adds the CWCMC and when it comes across a component that is set for cross wiring it sets it up

Correct, happy path. 

> or throws if the container isn't configured correctly

Let's validate against the configuration callback. Not whether the facility is registered.

>  the exception will wind up the stack through container.Register

That is exactly what we are trying to do. Like it. Thanks for the input.

*Summary*

We will move the validation down into the facility configuration extension. ie. if you call AddFacility without doing the config callback for `f => f.CrossWiresInto(IServiceCollection)` we throw helpfully. @generik0 I think this seems perfectly reasonable. We also have the documentation covered. What do you think?

generik0: @fir3pho3nixx I think this is fine. I have begun testing. It appears to be working as expected. Will reprot more ASAP.

fir3pho3nixx:
Welcome @xhafan! :)

This PR is now blocked by https://github.com/castleproject/Windsor/issues/411

src/Castle.Facilities.AspNetCore/AspNetCoreFacility.cs
fir3pho3nixx: @generik0 Sometimes it take me days/weeks. Don't stress. This is by far the worlds craziest PR. :)

The outstanding issue for this has been open since Feb 2, 2016. 

https://github.com/castleproject/Windsor/issues/120

Don't worry fella, take your time! :)

fir3pho3nixx: @jonorossi We OK to go ahead?

jonorossi: > We will move the validation down into the facility configuration extension. ie. if you call AddFacility without doing the config callback for f => f.CrossWiresInto(IServiceCollection) we throw helpfully. @generik0 I think this seems perfectly reasonable. We also have the documentation covered. What do you think?

It seems like you are agreeing with what I wrote but what I wrote isn't what you are doing, so I'm a little confused.

I thought you were going to make it so you always add the facility even if you don't use cross wiring, then the `ComponentRegistration` extension methods don't need to check any configuration because they aren't "attached" to a container yet, and only when you call `Register` with the `ComponentRegistration` does the container's configuration (i.e. that is has a `IServiceCollection` set) get checked?

fir3pho3nixx: > I thought you were going to make it so you always add the facility even if you don't use cross wiring

Correct. Once we established that, I was merely asking @generik0 if we move validation around the facility being present or not down a level of checking whether the IServiceCollection is assigned or not using a CWCMC as you suggested [here](https://github.com/castleproject/Windsor/pull/394#discussion_r194979053).  Just making sure we are all on the same page. 

> then the ComponentRegistration extension methods don't need to check any configuration because they aren't "attached" to a container yet.

Yes. You gave me a back reference to the IKernel [here](https://github.com/castleproject/Windsor/pull/394#discussion_r194979053). With `CrossWiringComponentModelContributor.ProcessModel` I can get at the kernel. Facility scan for `AspNetCoreFacility` and check whether IServiceCollection is null. Done. Remove all the validation from ComponentRegistration. Get rid of AsyncLocal.

Have I missed anything?

generik0:
Hi Guys.
Looks like we need to test the new generic-host also:
[generic-host](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-2.1)

fir3pho3nixx:
@generik0 are you sure? It looks to me like this is used in non-http hosting scenarios. They did this to decouple the http request pipeline. After reading this I imagine one would use this for long running tasks or ‘queue workers’.

fir3pho3nixx:
If we can converge with the DI for both scenarios that would be good. Although cautiously I think we need to raise a new issue to track this. 

Want to create one?

src/Castle.Facilities.AspNetCore/AspNetCoreFacility.cs
jonorossi: @fir3pho3nixx yes, I had wrongly misinterpreted your "We OK to go ahead?" with the changes were made and are we ready to go ahead and merge.

generik0:
@fir3pho3nixx you are correct. Its a new feature request, maybe a new facility.
Check: [Autofac Generic Host Example](https://github.com/autofac/Examples/tree/master/src/GenericHostBuilderExample)
This example uses the AutofacServiceProviderFactory. 

This may however lead to a new facility in castle that the aspnetcore also should use.?

src/Castle.Facilities.AspNetCore/AspNetCoreFacility.cs
fir3pho3nixx: Awesome let me get this going. Will implement it tonight. Thanks ;)

fir3pho3nixx:
@generik0 this sounds correct to me. Let me finish up this PR tonight and then we head over to the issue to flesh out the design a little more. 

Even if we end up going the long way round it might end up informing us that we don’t even need cross wiring anymore in this facility ... hoping! ;)

src/Castle.Facilities.AspNetCore.Tests/WindsorRegistrationExtensionsTestCase.cs
```diff
@@ -0,0 +1,354 @@
...
+			}
+		}
+
+		[Test]
+		public void Should_resolve_IMiddleware_from_Windsor()
+		{
+			testContext.ApplicationBuilder.UseMiddlewareFromWindsor<AnyMiddleware>(testContext.WindsorContainer);
+
+			Assert.DoesNotThrow(() => { testContext.WindsorContainer.Resolve<AnyMiddleware>(); });
+		}
```
fir3pho3nixx: Found more sucky API's. 

@jonorossi Do you have any tips here?

fir3pho3nixx: Would be great of we could get rid of the IWindsorContainer parameter here. It is a bit late so I will have to pick this up tomorrow.

jonorossi: Not knowing the ASP.NET Core stuff, I guess you'd have to store the container reference and get access via `IApplicationBuilder`.

fir3pho3nixx: @jonorossi 

Correct. I managed to take advantage of the facility to do this. Here is some sample code for new usage:

```csharp
public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
{
	// For making component registrations of middleware easier
	Container.GetFacility<AspNetCoreFacility>().RegistersMiddlewareInto(app);

	// Add custom middleware, do this if your middleware uses DI from Windsor
	Container.Register(Component.For<CustomMiddleware>().DependsOn(Dependency.OnValue<ILoggerFactory>(loggerFactory)).LifestyleScoped().AsMiddleware());
	
	// ... 
}
```

This now means we have first class support for Middleware registration(including lifestyles) using component registrations as opposed to the `roll your own` approach which coerced everything to singleton. 

I will present this change as a separate commit once I heard back from @xhafan on issue #411 to make the review a little easier and squish once we are all happy with the proposed change.

fir3pho3nixx: Man, sorting out the memory management was an absolute bitch. These were pre-existing problems where CrossWired components were consuming each other (https://github.com/fir3pho3nixx/Windsor/pull/25).

src/Castle.Facilities.AspNetCore/Activators/CrossWiredComponentActivator.cs
fir3pho3nixx: Please see: https://github.com/fir3pho3nixx/Windsor/pull/25#pullrequestreview-131406376

fir3pho3nixx: > I assume you weren't able to just remove the DisposalConcern from ComponentModel.Lifecycle in CrossWiringComponentModelContributor?

This has now also been done.

fir3pho3nixx:
LTGM, thanks to everyone that participated. Truly thank you! :)

stakx:
Sorry for not reviewing @fir3pho3nixx, GitHub never notified my of that requestand while I take a look at this PR every now and then I missed this one... sorry!

fir3pho3nixx:
Don’t worry bud! You are pretty busy with the core work. Not a problem ;)

PR #401, opened 2018-05-24
Update Facilities.md

GuillaumeSmartLiberty:
replace link to github project as it seems one the most active source compared to the previous dead link and other site http://sharparchitecture.github.io/ which looking a bit outdated


jonorossi:
Thanks

PR #406, opened 2018-06-02
Resolve #405. Unicode please!!!

### LINKED ISSUE #405: Some source code string literals are not utf-8
When trying to compile Windor on non-English platform, say Chinese, I got compilation errors that are caused by unrecognizable characters.

```
-			Assert.AreEqual("GenericHasNested< TOuter >.NestedGeneric<A2, AProp>[,,]", name);
+			Assert.AreEqual("GenericHasNested<·TOuter·>.NestedGeneric<A2, AProp>[,,]", name);
```

It turns out the file is not in utf-8 encoding. It would be nice if we save all source files in utf-8 encoding.

BTW, what is that `·` for?

davidshen84:
I converted that peculiar dot into utf-8 encoding, and the solution can build on my Windows 10 64 bit system which has **GB2312** as the default character encoding system.

Also, if possible, can we remove that dot?


fir3pho3nixx:
Hi David

Thanks for this. Do you think should run something like this over the entire repository? I tested it and it does not break anything tests on my side. 

```csharp
class Program
{
	static void Main(string[] args)
	{
		var inputDirectory = @"c:\code\windsor"; // <- change to your own checkout path

		WriteUTF8WithoutBOM(inputDirectory, "*.cs");
		WriteUTF8WithoutBOM(inputDirectory, "*.xml");
		WriteUTF8WithoutBOM(inputDirectory, "*.config");
		WriteUTF8WithoutBOM(inputDirectory, "*.csproj");
		WriteUTF8WithoutBOM(inputDirectory, "*.props");


		Console.ReadLine();
	}

	private static void WriteUTF8WithoutBOM(string inputDirectory, string searchPattern)
	{
		foreach (var file in Directory.GetFiles(inputDirectory, searchPattern, SearchOption.AllDirectories))
		{
			if (file.Contains(".git\\")) continue;
			if (file.Contains(".vs\\")) continue;

			Console.WriteLine(file);

			Encoding utf8WithoutBom = new UTF8Encoding(false);
			var fileContents = File.ReadAllText(file, utf8WithoutBom);
			var streamWriter = new StreamWriter(file, false, utf8WithoutBom);
			streamWriter.Write(fileContents);
			streamWriter.Dispose();
		}
	}
}
```

Also don't forget to add a CHANGELOG.md entry.

fir3pho3nixx:
> Also, if possible, can we remove that dot?

We don't have to remove the dot if the file is encoded properly.

davidshen84:
@fir3pho3nixx , good idea. I have updated the PR.

fir3pho3nixx:
@davidshen84 - Don't forget to make a note in the [CHANGELOG.md](https://github.com/castleproject/Windsor/blob/master/CHANGELOG.md)

jonorossi:
@fir3pho3nixx why did you want to remove BOMs? They are valid as part of the Unicode standard and not hurting anyone; Visual Studio is actually the one saving them there, not sure if the new versions of VS still adds them to new files.

I don't see a reason to mention this in the changelog, unless the change affects users of the library.

If I recall looking at the debugger views in Windsor the dots might be those fancy angle brackets, I recall seeing them somewhere in Windsor debugger views (added in 2011: d7ff1dabb6d9dc2cd239471fa728e979ec1510fb). I see no reason not to just make this function output exactly what you'd type as a C# developer, no need for fancy characters.

stakx:
If you need a good excuse to normalize BOMs across the whole codebase, you could just update the year to 2018 in the copyright... and do the BOM adjustment as an aside. ;-)

fir3pho3nixx:
Oh dear this was never about BOM’s. Sorry if that is what you thought. The code I supplied actually converts ANSI to UTF8 but happens to remove BOM’s. I thought this was a good approach as BOM’s infer things about character streams. 

@jonorossi This actually a defect which prevents people with non standard encoding from building the solution. So I thought it was a fix of some description. Updating the change log is neither here not there. As long as it builds! Sweet. 

@stakx I am not sure I follow. License headers and UTF encoding of source files that allow this solution to be built in Asia are very different. The real problem is we have some ANSI duff ruining the party.

stakx:
@fir3pho3nixx, please disregard my remark, I was half joking. (But to explain what I meant, I was also under the impression that this was about BOMs mainly, and since the diffs already affect the first line of each code file, which happens to contain the copyright line, you could've killed two birds with one stone but put the focus on only one of those. Sneaky, and not altogether serious.)

fir3pho3nixx:
@stakx - Awww man, if only the BOM's were distributed across all code files. Then we could update the headers! :)

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 Enhancements:
 - Added XML documentation to BeginScope and RequireScope lifetime extensions (@jonorossi)
 - Upgraded build to use NUnit Adapters (@fir3pho3nixx, #243)
+- Remove BOM from all source code files (@davidshen84, #405)
```
fir3pho3nixx: Maybe we should say:

`Fixed encoding problems prevented build under GB2312 character encoding system`

davidshen84:
Yeah~the discussions carried me away, and I forgot why I opened this bug in
the first place :)

stakx:
> Also, if possible, can we remove that dot?

Including this the PR referenced above.

fir3pho3nixx:
Interesting to note we have 624 files with BOM's and 1508 without BOM's.

fir3pho3nixx:
Do we care about the endian nature of our code?

fir3pho3nixx:
> What endianness are you referring to? UTF-8 does not have any endianness, it's a byte stream. Endianness would only be applicable with UTF-16 or UTF-32 (AFAIK).

It infers endian nature for character streams. 
https://en.wikipedia.org/wiki/Byte_order_mark

stakx:
> It infers endian nature for character streams.
> https://en.wikipedia.org/wiki/Byte_order_mark

I am still not sure I understand what you're referring to. Unlike with UTF-16 (where the BOM can be the byte sequence `fe ff` or `ff fe`, depending on endianness), there is no such thing as Little Endian UTF-8 or Big Endian UTF-8. There's only UTF-8, and its BOM is invariably the byte sequence `ef bb bf`. So unless you're converting everything to UTF-16, the question of endianness simply should not arise...?

fir3pho3nixx:
Yes and with UTF-8 it is 0xEF,0xBB,0xBF.

stakx:
As to the question of whether to reverse the BOM change or not: I was just passing by, so I'll leave the decision up to you folks. (My purely personal opinion is that the the IT industry has got to a point where tools can finally be expected to handle UTF-8 properly. We should no longer have to adjust to broken tooling, so removing the BOM for that reason seems like the wrong incentive. The broken tooling should be made UTF-ready instead. On a second note, the presence of a UTF-8 BOM can help tools recognize UTF-8 encoding, but today most good tooling probably assumes that encoding by default, anyway. So whether the BOM is present or not shouldn't make much difference, *as long as* the characters in the file are actually encoded as UTF-8. I agree that from an "aesthetic" viewpoint, it feels nice to have uniformity, but it's probably not worth spending too much time on the matter. It's fine either way.)

fir3pho3nixx:
> My purely personal opinion is that the the IT industry has got to a point where tools can finally be expected to handle UTF-8 properly. We should no longer have to adjust to broken tooling, so removing the BOM for that reason seems like the wrong incentive. The broken tooling should be made UTF-ready instead.

Cool. This PR just grew to a change set over 624 files to one of 1508 files. 

> On a second note, the presence of a UTF-8 BOM can help tools recognize UTF-8 encoding, but today most good tooling probably assumes that encoding by default, anyway. So whether the BOM is present or not shouldn't make much difference, as long as the characters in the file are actually encoded as UTF-8. I agree that from an "aesthetic" viewpoint, it feels nice to have uniformity, but it's probably not worth spending too much time on the matter. It's fine either way.)

Pick one.

fir3pho3nixx:
Please can we just merge this.

fir3pho3nixx:
@davidshen84 - Can you please merge my PR into your master branch.

davidshen84:
@fir3pho3nixx, merged. I think we should have opened another issue to discuss this BOM problem. This issue was only about a few files that were in ANSI encoding.

fir3pho3nixx:
> @fir3pho3nixx, merged. I think we should have opened another issue to discuss this BOM problem. 

True. I just felt encoding was a valid way to bring this up.

> This issue was only about a few files that were in ANSI encoding.

1 from what I could tell.

fir3pho3nixx:
Sorry for all the stress people!

stakx:
@fir3pho3nixx - Please accept my apologies if any of my above statements upset you... that certainly wasn't my intention at all! While I don't see any immediate need for it, I can still appreciate why you'd like to harmonize BOMs across the codebase. I'll stay out of this. Again, sorry if I upset you.

fir3pho3nixx:
> Please accept my apologies if any of my above statements upset you...

Never. You can tell me off anytime! :)

> I'll stay out of this. Again, sorry if I upset you.

No. You are always welcome.

jonorossi:
> I think we should have opened another issue to discuss this BOM problem. This issue was only about a few files that were in ANSI encoding.

Oops, yer this thread appears to have gone a bit off track. Looking at your commit that matters (62e2461dac80b369f652ae56f2a63c08141c38f6), @stakx has made this fix in #407 along with others to `TypeUtil.ToCSharpString` which was a little broken so we can close this pull request without merging. Thanks for bringing up this problem.

I've logged #408 to continue the BOM discussion.

fir3pho3nixx:
Sorry sorry sorry

PR #407, opened 2018-06-05
Make Typeutil.ToCSharpString produce more C#-like formatting

### LINKED ISSUE #404: TypeUtil.ToCSharpString(this Type) does not handle nested generic types correctly
_Migrated from https://github.com/castleproject/Core/issues/365._

---

_@pmg23 wrote:_

The useful extension method `TypeUtil.ToCSharpString(this Type)` does not correctly deal with types nested within generic types [as of Core 4.2.1]. For example:

```cs
internal class A<T1>
{
    internal class B {}
    internal class C<T2> {}
}

Assert.That(typeof(A<string>.B).ToCSharpString(),
        Is.EqualTo("A<String>.B")); // FAIL: actually "A<·T1·>.B<String>"
Assert.That(typeof(A<string>.C<int>).ToCSharpString(),
        Is.EqualTo("A<String>.C<Int32>")); // FAIL: actually "A<·T1·>.C<String, Int32>"
```

---

_@stakx replied:_

(Shameless advertisement:)

Type name formatting is surprisingly complex because there are quite a few corner cases to deal with. I happen to have written a small library recently ([stakx/TypeNameFormatter](https://github.com/stakx/TypeNameFormatter)) that would solve this problem. It's available as a source code NuGet package.

---

_@stakx replied:_

That being said, I cannot seem to find a method `TypeUtil.ToCSharpString(this Type)` in Castle Core (this repository). Did you mean [this method](https://github.com/castleproject/Windsor/blob/a9d6ad8c7ad8d343d2017c97a67d6e1c2fb36c39/src/Castle.Windsor/Core/Internal/TypeUtil.cs#L50-L63) in Castle Windsor ([castleproject/Windsor](https://github.com/castleproject/Windsor))?

stakx:
This changes the way how `TypeUtil.ToCSharpString` formats certain types to a representation that looks more like C#:

* **Generic type parameters.** Those are currently surrounded with a centered dot (`·`) that noone quite seems to know what they're good for. <sup>(&dagger;)</sup> (See #406, esp. the recommendation made in https://github.com/castleproject/Windsor/pull/406#issuecomment-394178922.)

* **Nested generic types.** For example, given `Outer<bool>.Inner<char>`, it outputs `"Outer<·T·>.Inner<bool, char>"` which is likely to be confusing rather than helpful. This changes the way how such types get formatted to something more familiar to C# users: `"Outer<bool>.Inner<char>"`. (This resolves #404.)

These changes are classified in the changelog as an "enhancement". Given that there were explicit unit tests about that behavior, it's not really a bugfix, but a change. Classifying as a "breaking change" would be accurate, but given that the method's intended use case is for diagnostics, it seems somewhat over-the-top to mark it as "breaking".

<sup>(&dagger;)</sup> My personal theory about the centered dots is that they were needed to mark a placeholder, so you'd know how to apply, or "distribute", the generic type arguments from the innermost type over the outer types. Now that this distribution is done automatically, the dots aren't really needed anymore.

### DIFF
--- a/src/Castle.Windsor.Tests/TypeUtilTestCase.cs
+++ b/src/Castle.Windsor.Tests/TypeUtilTestCase.cs
@@ -14,6 +14,8 @@
 
 namespace CastleTests
 {
+	using System;
+
 	using Castle.ClassComponents;
 	using Castle.Core.Internal;
 	using Castle.MicroKernel.Tests.ClassComponents;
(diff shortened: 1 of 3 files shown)


src/Castle.Windsor.Tests/TypeUtilTestCase.cs
```diff
@@ -55,38 +55,38 @@ public void Closed_generic_simple_type()
...
+			Assert.AreEqual("GenericHasNested<A2>.Nested[]", name);
 		}
 
 		[Test]
-		public void Generic_nested_type_ignores_outer_generic_argument()
+		public void Generic_nested_type()
 		{
 			var name = typeof(GenericHasNested<A2>.Nested).ToCSharpString();
-			Assert.AreEqual("GenericHasNested<�TOuter�>.Nested<A2>", name);
+			Assert.AreEqual("GenericHasNested<A2>.Nested", name);
```
fir3pho3nixx: Should we add a test that nests generic arguments? 

```csharp
var name = typeof(GenericHasNested<A2>.GenericNested<Tuple<int, bool>>).ToCSharpString();
Assert.AreEqual("GenericHasNested<A2>.GenericNested<Tuple<int, bool>>", name);
```

fir3pho3nixx:
@stakx beat me to it! :)

I had mine on a branch but did not finish it in time.

src/Castle.Windsor.Tests/TypeUtilTestCase.cs
stakx: Will do. It's good to have a test to see whether the recursive type forwarding happens only where it should.

Beyond that, I would caution against blowing up this utility part of Windsor too much. If we wanted a really robust implementation for type name formatting, I'd be happy to donate my own recent work (stalx/TypeNameFormatter) which comes tested and as a source code NuGet package that easily embeddable (in the spirit of not reinventing the wheel and doing double work). That said, I perfectly understand if Castle would rather stay free of 3rd-party code.

fir3pho3nixx: I did have a close look at that, it is mental the acrobatics one has to do to achieve this. I would like to avoid a third party dependency on this but ultimately it is for @jonorossi to decide. Thanks for doing this. The last test case you added was where I got up to :)

stakx: I've all done it before so it wasn't too bad this rime around.

Try a type such as `int[][,][,,]` and see what happens to the commas. ;-)

Totally fine with *not* introducing a dependency, I just thought I'd mention it as a possibility.

jonorossi: Thanks for the offer @stakx, I don't mind the contribution though would prefer it to be embedded rather than fetched from NuGet, to be embedded it probably would also be best licensed under Apache 2. However I think what we've now got covers what the debugger views are going to have to worry about anyway, I can't imagine any user of Windsor having a component of type `int[][,][,,]` (can't imagine Windsor resolving it 😉), so for pragmatism sake I think your fix here is all we need.

jonorossi:
Merged, thanks.

src/Castle.Windsor.Tests/TypeUtilTestCase.cs
stakx: >  I can't imagine any user of Windsor having a component of type int[][,][,,]

You'd be surprised what contraptions people come up with when you least expect it! :laugh:

> so for pragmatism sake I think your fix here is all we need.

Agreed, let's just wait and see if anyone complains about this again.

fir3pho3nixx: > You'd be surprised what contraptions people come up with when you least expect it! 😆

Duuuude. I feel you on that.

Issue #423, opened 2018-07-30
WindsorRegistrationExtensions.AddApplicationComponentsToWindsor


MAliM1988:
Hi Everyone

in this Method I Have Problem For Registering `ControllerBase` For Registering Api Controller in asp Net Core 2.1

in this Method you Use this Code 

```
foreach (var controllerRegistration in options.ControllerRegistrations)
{
	container.Register(Classes.FromAssemblyInThisApplication(controllerRegistration.Item1).BasedOn<Controller>().Configure(x => x.LifeStyle.Is(controllerRegistration.Item2)));
}
```

And Ignore `ControllerBase` Type,I Use this Code For Fix this Problem

```
foreach (var controllerRegistration in options.ControllerRegistrations)
{
	//container.Register(Classes.FromAssemblyInThisApplication(controllerRegistration.Item1).BasedOn<Controller>().Configure(x => x.LifeStyle.Is(controllerRegistration.Item2)));
	container.Register(Classes.FromAssemblyInThisApplication(controllerRegistration.Item1).BasedOn<ControllerBase>().Configure(x => x.LifeStyle.Is(controllerRegistration.Item2)));
}
```

fir3pho3nixx:
https://github.com/castleproject/Windsor/blob/master/docs/aspnetcore-facility.md#registering-controllers-viewcomponents-and-taghelpers

```csharp
services.AddWindsor(container, opts =>
{
	opts.RegisterControllers(typeof(HomeController).Assembly, LifestyleType.Transient);
	opts.RegisterTagHelpers(typeof(EmailTagHelper).Assembly, LifestyleType.Transient);
	opts.RegisterViewComponents(typeof(AddressComponent).Assembly, LifestyleType.Transient);
});
```
[WindsorRegistrationOptions](https://github.com/castleproject/Windsor/blob/master/src/Castle.Facilities.AspNetCore/WindsorRegistrationOptions.cs#L25) was meant to be the way of overriding defaults used for registration. Do you think we should expose the [component registration](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/Component.cs#L33) api's here so we gain maximum flexibility?

Something like:

 ```csharp
services.AddWindsor(container, opts =>
{
	opts.RegisterControllers(Classes.FromAssemblyInThisApplication(options.EntryAssembly).BasedOn<ControllerBase>().LifestyleScoped());
});
```

[...]

Issue #415, opened 2018-06-29
Delegate factory with Dependency.OnComponent and TypedFactoryFacility


Qwetek:
Hello, i have been doing some refactoring of our container mess and it seems i found something weird. When using delegate factory and Dependecny.OnComponent at once the factory gets wrong component resolved, i boiled it down to this code:
```c#
public interface IInterface {}
public class C1 : IInterface {}
public class C2 : IInterface {}
public  class Root
{
  private IInterface _I;
  public Root(Func<IInterface> factory)
    {
	  //here i expect to get instance of C2 but get C1
      _I = factory();
    }
}

public void Main()
{
  _container = new WindsorContainer();
  _container.AddFacility<TypedFactoryFacility>();
  
  _container.Register(
  	Component.For<IInterface>().ImplementedBy<C1>(),
  	Component.For<IInterface>().ImplementedBy<C2>(),
  	Component.For<Root>().DependsOn(Dependency.OnComponent<IInterface, C2>())
  );
  
  var t = _container.Resolve<Root>();
}
```

I used a workaround with DependencyOnValue and created the delegate factory myself, but thats ugly:

`Component.For<Root>().DependsOn(Dependency.OnValue<Func<IInterface>>((Func<IInterface>)(_ => new C2())`

So am i doing something wrong ? Or is it a bug ? Or maybe its by design ? Is there a better way than the Dependency.OnValue ?
Note: this example is verry simplified and sadly refactoring the classes is out of the question.
Note2: the delgate (Func<IInterface>) in my case takes arguments, i just removed them because they are not neccesary for the example

varsion v4.0.30319 (i know its not the newest, but i work with what i have and if neccesary i can upgrade )

fir3pho3nixx:
Ok, when you register IInterface twice with two components C1 and C2 what exactly are you expecting to happen via the dependency resolve graph? 

Does the C1 registration supersede C2 or vice versa? What happens with a IInterface dependency? Do we need an array resolver? 

Your registration code is a little confusing. Sorry!

fir3pho3nixx:
For example, ASP.NET core DI will let you override registrations, Windsor does not.

fir3pho3nixx:
@jonorossi 

Do you know the history behind the `DependsOn` component registration API? I can clearly see this was an attempt at washing over the service override's API using strong typing. What do we want to do with this? It seems like abstraction for abstraction sake, and as @Qwetek highlighted there are clearly bug with this.

[...]

PR #424, opened 2018-08-03
Registration Type in WindsorRegistrationOptions 

MAliM1988:
Issues [423](https://github.com/castleproject/Windsor/issues/423)

[WindsorRegistrationOptions ](https://github.com/castleproject/Windsor/blob/master/src/Castle.Facilities.AspNetCore/WindsorRegistrationOptions.cs#L25) was meant to be the way of overriding defaults used for registration.  should expose the [component registration](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/MicroKernel/Registration/Component.cs#L33) api's for maximum flexibility

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,15 @@
 ## Unreleased
 
 Enhancements:
-- Upgraded to Castle.Core 4.2.0 to 4.3.0. https://github.com/castleproject/Core/releases/tag/v4.3.0 (@fir3pho3nixx).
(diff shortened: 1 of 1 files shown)


fir3pho3nixx:
@MAliM1988 

I think maybe there was a miscommunication, these commits are for issue #415. This should also not be going to master. 

Let me raise a PR and tag you to show you what I meant. One sec.

fir3pho3nixx:
@MAliM1988 

Please see my comment here with a demonstration in code of the suggested change: https://github.com/castleproject/Windsor/issues/423#issuecomment-410339741

PR #420, opened 2018-07-14
Deprecating `Resolve(Type, IDictionary)` in favour of `Resolve(Type, IReadOnlyDictionary<string, object>)`

fir3pho3nixx:
Resolves: #346

This was implemented at the upper most surface API's for IWindsorContainer and WindsorContainer.

Once the overload `Resolve(Type, IDictionary)` was renamed to `Resolve(Type, IReadOnlyDictionary<string, object>)`, the compiler then redirected all old calls from `Resolve(Type, IDictionary)` to `Resolve(Type, object)`. This caused quite a few tests to fail, it also required a bit of type checking and casting for the resolve methods in the WindsorContainer to make sure they got to the right overload in the MicroKernel. After that tests started passing again. 

We should definitely revisit this at some point.

PR #426, opened 2018-08-06
Deprecate Lock in favour of ReaderWriterLockSlim

fir3pho3nixx:
As part of https://github.com/castleproject/Core/issues/376 the change for removing [Lock](https://github.com/castleproject/Core/pull/391/files#diff-38265397b99a6db95030e27b3f143441L26) in Castle.Core has left us wanting for new ways down here in Windsor.  

This is an experimental as per https://github.com/castleproject/Core/issues/376#issuecomment-410371303

### DIFF
--- a/src/Castle.Windsor/MicroKernel/Releasers/LifecycledComponentsReleasePolicy.cs
+++ b/src/Castle.Windsor/MicroKernel/Releasers/LifecycledComponentsReleasePolicy.cs
@@ -40,7 +40,7 @@ namespace Castle.MicroKernel.Releasers
 		private readonly Dictionary<object, Burden> instance2Burden =
 			new Dictionary<object, Burden>(ReferenceEqualityComparer<object>.Instance);
 
-		private readonly Lock @lock = Lock.Create();
+		private readonly ReaderWriterLockSlim @lock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
 		private readonly ITrackedComponentsPerformanceCounter perfCounter;
 		private ITrackedComponentsDiagnostic trackedComponentsDiagnostic;
 
@@ -84,16 +84,16 @@ namespace Castle.MicroKernel.Releasers
 		{
 			get
 			{
-				using (var holder = @lock.ForReading(false))
+				@lock.EnterReadLock();
+
+				try
 				{
-					if (holder.LockAcquired == false)
-					{
-						// TODO: that's sad... perhaps we should have waited...? But what do we do now? We're in the debugger. If some thread is keeping the lock
-						// we could wait indefinatelly. I guess the best way to proceed is to add a 200ms timepout to accquire the lock, and if not succeeded
-						// assume that the other thread just waits and is not going anywhere and go ahead and read this anyway...
-					}
 					var array = instance2Burden.Values.ToArray();
 					return array;
+				}
+				finally
+				{
+					@lock.ExitReadLock();
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


fir3pho3nixx:
@stakx Can you have a look? Will update commit comments before we merge.

fir3pho3nixx:
Oh and changelog of course

src/Castle.Windsor/MicroKernel/Releasers/LifecycledComponentsReleasePolicy.cs
```diff
@@ -84,33 +84,43 @@ private Burden[] TrackedObjects
 		{
 			get
 			{
-				using (var holder = @lock.ForReading(false))
+				@lock.EnterReadLock();
```
stakx: Not sure if you want to preserve exact semantics here (probably not, since you're removing the below TODO)... but if you do, `TryEnterReadLock` would be the right method to call instead.

src/Castle.Windsor/MicroKernel/Releasers/LifecycledComponentsReleasePolicy.cs
```diff
@@ -147,7 +164,10 @@ public void Release(object instance)
 			}
 
 			Burden burden;
-			using (@lock.ForWriting())
+
+			@lock.EnterWriteLock();
```
stakx: As a 1:1 conversion, this looks good. But I can't help feeling a bit confused about what is going on here. This code is taking a write lock when it's only reading a dictionary. The `Release` call (which I assume triggers said `OnInstanceReleased` event?) happens outside the lock. Should this probably be an upgradeable read lock, with the call to `Release` being moved inside it... or am I misunderstanding how this code works?

src/Castle.Windsor/MicroKernel/Releasers/LifecycledComponentsReleasePolicy.cs
```diff
@@ -169,12 +194,12 @@ public virtual void Track(object instance, Burden burden)
...
 						instance, lifestyle));
 			}
+
 			try
 			{
-				using (@lock.ForWriting())
-				{
-					instance2Burden.Add(instance, burden);
-				}
+				@lock.EnterWriteLock();
```
stakx: In other places, the code entering the lock is placed outside the `try`. Why not here?

src/Castle.Windsor/MicroKernel/Releasers/LifecycledComponentsReleasePolicy.cs
```diff
@@ -185,19 +210,31 @@ public virtual void Track(object instance, Burden burden)
...
 			burden.Released += OnInstanceReleased;
 			perfCounter.IncrementTrackedInstancesCount();
 		}
 
 		private void OnInstanceReleased(Burden burden)
 		{
-			using (@lock.ForWriting())
+			try
 			{
+				@lock.EnterWriteLock();
```
stakx: Same here, normally the lock entering code is placed before the `try`, but not here.

stakx: Also, this (`OnInstanceReleased` taking a write lock) appears to support my suspicion that `Release` (above) should perhaps take an upgradeable read lock and do the `Release` call inside it.

src/Castle.Windsor/MicroKernel/Releasers/LifecycledComponentsReleasePolicy.cs
```diff
@@ -84,33 +84,43 @@ private Burden[] TrackedObjects
...
 				}
+
 				burdens = instance2Burden.ToArray();
 				instance2Burden.Clear();
 			}
+			finally
+			{
+				@lock.ExitWriteLock();
+			}
+
```
stakx: A note about the `foreach` loop following this code. It calls `Release` on each `burden` in the `burdens` collection. Each `Release` will take a write lock. That seems kind of inefficient to me. If I've read this correctly, then this perhaps could be more efficient, something along the lines of:

1. Move the actual logic of `Release` (all except the lock taking) into a new, separate method `ReleaseWithoutTakingLock`.
2. Delegate to that new method from inside `Release`.
3. Before the `foreach` loop (here), take a lock. Inside the loop, use the new non-locking method.

I could be wrong, as I haven't studied this code in detail. But perhaps worth giving a thought or two.

fir3pho3nixx:
My give up.

stakx:
Hi @fir3pho3nixx - just saw that you closed this. I'd quickly like to enquire whether the reason for this  might have been any insurmountable problems you stumbled upon while migrating Windsor away from Core's `Lock` API? If so, that'd be good to know, as it might mean we need to revert the `Lock` API's deprecation in Castle in order to not leave Windsor behind. Cheers for a brief feedback! :)

fir3pho3nixx:
What happened here was nothing to do with the locks at all. I just really now feel we should release ASP.NET Core for Windsor. I am making a bold statement by closing out my PR's until that is the case. I tried #420 but I think it is becoming unreasonable by asking questions about call vs callvirt.

fir3pho3nixx:
I hope to re-open this PR once ASP.NET Core is released.

Issue #110, opened 2015-12-08
TypedFacilityFactory component selection does not work with interfaces


Rahkola:
I'm trying to install an interface as an factory and then override the component selector class so that I would have separate component selector classes for different projects in my solution.

container.Register(
            Component.For<IMySelector>().ImplementedBy<ASelector>()
            Component.For<IMyFactory>().AsFactory(f => f.SelectedWith<IMySelector>());

And then on other project I'm trying to override the component selector:

container.Register(
            Component.For<IMySelector>().ImplementedBy<BSelector>());

This causes the factory instance to be null at runtime. Seems like the component selector is not resolved and needs to be a concrete class. Interface should also be supported here. If I use concrete class for the SelectedWith parameter, then it works:

container.Register(Component.For<IMyFactory>().AsFactory(f => f.SelectedWith<ASelector>());

mario-d-s:
I'm kind of confused by this issue because of the "pseudo" code. @Rahkola are you still around for clarifications?

I just had a quick look in the code. My first understanding is that the facility *should* already be resolving component selectors from the container, so this might just be a bug.

I'll try to raise a PR with a failing test for the case as originally described.

mario-d-s:
> This causes the factory instance to be null at runtime.

This is weird in its own right, as the test I've just written throws an exception saying "The referenced component {ISelector} could not be resolved". To end up with `null`, something else must be going on in the original scenario.

Rahkola:
Hi,

I'm still around :) This happened in somewhat more complex enterprise app that I was developing. Can't recall the actual case anymore, but I can try to re-produce the issue with more isolated environment.  It may be that the problem was caused by some other reason as you are suggesting.

I'm a bit busy with other projects at the time, but Ill try to find time to provide you isolated code which re-produces the issue. No promises though... If I don't come back to you in couple of weeks with the example code the ticket can be closed on by behalf.

PR #429, opened 2018-08-22
Add failing test for issue #110

mario-d-s:
I'm not sure if I've chosen the right file to put it in. Also unsure about the nested interface/class pair or their naming. @fir3pho3nixx would be glad if you could review.

### DIFF
--- a/src/Castle.Windsor.Tests/Facilities/TypedFactory/TypedFactorySelectorsTestCase.cs
+++ b/src/Castle.Windsor.Tests/Facilities/TypedFactory/TypedFactorySelectorsTestCase.cs
@@ -59,7 +59,7 @@ namespace CastleTests.Facilities.TypedFactory
 
 		}
 
-		[Test, Explicit("Failing test case for issue #110")]
+		[Test]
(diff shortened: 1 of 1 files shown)


src/Castle.Windsor.Tests/Facilities/TypedFactory/TypedFactorySelectorsTestCase.cs
```diff
@@ -58,5 +58,26 @@ public void Implicitly_specified_name_falls_back_if_not_present()
 			Assert.Throws<ComponentNotFoundException>(() => factory.Create());
 
 		}
+
+		[Test, Explicit("Failing test case for issue #110")]
```
fir3pho3nixx: Small nit. Can we please remove the explicit attribute? I would like these tests to fail outright in any builds.

mario-d-s: Sure thing, please check ed15d30.

fir3pho3nixx:
Thanks

PR #439, opened 2018-10-06
Typed factory burden on unrelated object

### LINKED ISSUE #436: Windsor adds a dependency resolved from typed factory as a burden on an unrelated object
Windsor 4.1.0 adds a dependency resolved from typed factory as a burden on an unrelated object. When that unrelated object is released, the dependency is disposed even though it's still in use by the object that actually resolved it.

This had been causing some hair-pulling all afternoon; sorry to bring another of these upon you. 😃

There's repro code at the end. Here's the clearest description I can write, besides the code itself:

➡ When a short-lived object depends on both a typed factory and a long-lived object which itself depends on the same typed factory, and the short-lived object’s constructor triggers the long-lived object to resolve a dependency of its own through the typed factory, Castle Windsor erroneously adds that dependency as a child burden to be disposed when the short-lived object is released. Even though that dependency was resolved by the long-lived service and the short-lived object does not know about it!

(In the real scenario, neither the long-lived or short-lived objects directly depend on the typed factory but rather they both depend on the same intermediate dependency which depends on the typed factory. I was able to remove the intermediate dependency and reproduce the same behavior.)

The debugger shows that the SqlConnection which the LongLivedService resolves is added as a burden to ShortLivedViewModel (breakpoint in [`Burden.AddChild`](https://github.com/castleproject/Windsor/blob/07e4b667c1826decdba1724642d6d68c2080c64e/src/Castle.Windsor/MicroKernel/Burden.cs#L97)).

-----

Self-contained repro:

##### Program.cs

```c#
using System;
using System.Data.SqlClient;
using Castle.Facilities.TypedFactory;
using Castle.Windsor;

public static class Program
{
    public static void Main()
    {
        using (var container = new WindsorContainer())
        {
            container.AddFacility<TypedFactoryFacility>();
            container.Register(
                Component.For(typeof(IFactory<>)).AsFactory(),
                Component.For<SqlConnection>().UsingFactoryMethod(kernel => new SqlConnection()).LifeStyle.Transient,
                Component.For<LongLivedService>().LifeStyle.Singleton,
                Component.For<ShortLivedViewModel>().LifeStyle.Transient);

            var longLivedService = container.Resolve<LongLivedService>();
            var shortLivedViewModel = container.Resolve<ShortLivedViewModel>();

            var prematureDisposalHandler = new EventHandler((s, e) =>
                throw new Exception("Long-lived service’s connection was disposed when short-lived view model was released."));

            longLivedService.SqlConnection.Disposed += prematureDisposalHandler;
            container.Release(shortLivedViewModel);
            longLivedService.SqlConnection.Disposed -= prematureDisposalHandler;

            container.Release(longLivedService);
            Console.WriteLine("All good!"); // Is never reached
        }
    }
}

public interface IFactory<T>
{
    T Resolve();
    void Release(T instance);
}

public sealed class LongLivedService
{
    public IFactory<SqlConnection> ConnectionFactory { get; }

    public SqlConnection SqlConnection { get; private set; }

    public LongLivedService(IFactory<SqlConnection> connectionFactory)
    {
        ConnectionFactory = connectionFactory;
    }

    public void StartSomething()
    {
        SqlConnection = ConnectionFactory.Resolve();
    }

    public void Dispose()
    {
        ConnectionFactory.Release(SqlConnection);
    }
}

public sealed class ShortLivedViewModel
{
    public IFactory<SqlConnection> ConnectionFactory { get; }
    public LongLivedService LongLivedService { get; }

    public ShortLivedViewModel(IFactory<SqlConnection> connectionFactory, LongLivedService longLivedService)
    {
        ConnectionFactory = connectionFactory;
        LongLivedService = longLivedService;
        longLivedService.StartSomething();
    }
}

```

##### CastleWindsorRepro.csproj

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net472</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Castle.Windsor" Version="4.1.0" />
  </ItemGroup>

</Project>
```

jnm2:
Fixes #436. The test I added now passes, the 13 tests that fail on master still fail, and all the tests that pass on master still pass.

@jonorossi I'm pretty confident that this is the right change to make. The fact that the factory is resolving in the call stack of the ShortLivedViewModel constructor call is accidental and Windsor should be acting the same as if the factory was resolving the component after a short delay (no longer in a component's constructor call stack). I added a comment to document this. Do you agree with this assessment?

A 4.1.1 release would be nice but in the meantime we'll gladly use the AppVeyor feed!


src/Castle.Windsor/MicroKernel/IKernelInternal.cs
```diff
@@ -51,7 +51,7 @@ public interface IKernelInternal : IKernel
 
 		IDisposable OptimizeDependencyResolution();
 
-		object Resolve(Type service, IDictionary arguments, IReleasePolicy policy);
+		object Resolve(Type service, IDictionary arguments, IReleasePolicy policy, bool ignoreParentContext = false);
```
jnm2: Adding this parameter is a breaking change to implementors of `IKernelInternal`. Given the name, I'm assuming users are expected to _not_ be implementing this interface.

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -735,12 +735,17 @@ protected virtual void RegisterSubSystems()
 		}
 
 		protected object ResolveComponent(IHandler handler, Type service, IDictionary additionalArguments, IReleasePolicy policy)
+		{
+			return ResolveComponent(handler, service, additionalArguments, policy, ignoreParentContext: false);
+		}
+
+		private object ResolveComponent(IHandler handler, Type service, IDictionary additionalArguments, IReleasePolicy policy, bool ignoreParentContext)
```
jnm2: The `protected` method above can be deleted if you don't think users will have inherited from `DefaultKernel` and called `ResolveComponent(IHandler, Type, IDictionary, IReleasePolicy)` from inside.

jonorossi:
Sorry for the delay, I'll check this out tomorrow.

jnm2:
Should I start a 'fixes' section in the changelog?

jnm2:
Started a 'fixes' section in the changelog and removed whitespace trimming per @stakx's feedback on my other PR.

src/Castle.Windsor/MicroKernel/IKernelInternal.cs
jonorossi: > I'm assuming users are expected to _not_ be implementing this interface.

Correct, it is there because Castle used to have two containers, the MicroKernel and Windsor, the two got merged but obviously still visible from the code base.

jonorossi:
> I'm pretty confident that this is the right change to make. The fact that the factory is resolving in the call stack of the ShortLivedViewModel constructor call is accidental. Windsor should be acting the same as if the factory was resolving the component after a short delay (no longer in a component's constructor call stack). I added a comment to document this. Do you agree with this assessment?

Yes, this appears a like a sound change, I'm not familiar enough with the typed factory code to know if this could be done a little cleaner, but I'm happy for the change to go in. At least we've got a unit test for it now, and it knocks one typed factory defect off.
 
> A 4.1.1 release would be nice but in the meantime we'll gladly use the AppVeyor feed!

I've created a 4.1 branch off v4.1.0, switch this pull request over to that base and we can get the change out. I'll get the change merged back into master.

jonorossi:
> [...] removed whitespace trimming per @stakx's feedback on my other PR.

I'm using `trim_trailing_whitespace = true` in my current project's editorconfig, if we added it with false would your editor behave?

jnm2:
> I'm using `trim_trailing_whitespace = true` in my current project's editorconfig, if we added it with false would your editor behave?

I know VS code would. I'd have to check on ReSharper/VS.

jnm2:
Switched branches.

Re: how clean of a change it is. I don't know either. My rationale is that the per-call-stack tracking is not appropriate (it's the cause of the problem) and that this is the simplest way to turn it off when it's not appropriate.

jonorossi:
> Switched branches.

You'll need to rebase too, otherwise this pull request will drag 61 commits from master across.

jnm2:
@jonorossi Done, sorry! Did I update the changelog properly?

robsonj:
I'm trying to upgrade from Castle.Windsor 4.1.0 to 4.1.1 and seeing a unit test failure between the two versions, the second assert fails in the following unit test for 4.1.1, is this expected behavior?


[WindsorContainerTypedFactoryTests.cs.txt](https://github.com/castleproject/Windsor/files/2534780/WindsorContainerTypedFactoryTests.cs.txt)

jonorossi:
@robsonj could you please create a new issue referencing this pull request.

robsonj:
Opened Issue #451, thank you.

PR #440, opened 2018-10-11
Fix debug mode test failures

### LINKED ISSUE #421: Test fail when run using Visual Studio 15.7.5 and Resharper 2018.1.3 but not from build.cmd for net45
When running test inside Visual Studio using Resharper I noticed that 13 tests are failing. However when I run the test from `build.cmd` everything passes. I just updated my Visual Studio and Resharper today and noticed this to be the case. 

image

image

jnm2:
Fixes #421. ~~The first commit isn't related (see commit message).~~ (now moved to #441)

There are now no occurrences of the text `GC.Collect` except in `ReferenceTracker.cs`, and no occurrences of the text `weak` (case insensitive) except in `ReferenceTracker.cs` and [`ModelFakes.cs`](https://github.com/castleproject/Windsor/blob/07e4b667c1826decdba1724642d6d68c2080c64e/src/Castle.Facilities.AspNetCore.Tests/Fakes/ModelFakes.cs#L38) (looks unrelated).

Hopefully the new test helper makes the tests easier to follow than they were before?

### DIFF
--- a/src/Castle.Facilities.WcfIntegration.Tests/Castle.Facilities.WcfIntegration.Tests.csproj
+++ b/src/Castle.Facilities.WcfIntegration.Tests/Castle.Facilities.WcfIntegration.Tests.csproj
@@ -24,6 +24,7 @@
 		<ProjectReference Include="..\Castle.Facilities.Logging\Castle.Facilities.Logging.csproj" />
 		<ProjectReference Include="..\Castle.Facilities.WcfIntegration.Demo\Castle.Facilities.WcfIntegration.Demo.csproj" />
 		<ProjectReference Include="..\Castle.Facilities.WcfIntegration\Castle.Facilities.WcfIntegration.csproj" />
+		<ProjectReference Include="..\Castle.Windsor.Tests\Castle.Windsor.Tests.csproj" />
 		<ProjectReference Include="..\Castle.Windsor\Castle.Windsor.csproj" />
 	</ItemGroup>
 
--- a/src/Castle.Windsor.Tests/ReferenceTracker.cs
+++ b/src/Castle.Windsor.Tests/ReferenceTracker.cs
@@ -1,0 +1,137 @@
+﻿// Copyright 2018 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace CastleTests
+{
+	using System;
+
+	using NUnit.Framework;
+
+	public static class ReferenceTracker
+	{
+		/// <summary>
+		/// <para>
+		/// Use a lambda method to evaluate an expression returning an object instance to track.
+		/// </para>
+		/// <para>
+		/// (If the expression is evaluated directly inside the test method, the runtime will keep it alive
+		/// until the test method returns if the runtime is in debug mode. ReSharper’s test runner runs in
+		/// debug mode when the active solution configuration is Debug.)
+		/// </para>
+		/// </summary>
+		public static ReferenceTracker<T> Track<T>(Func<T> getInstanceToTrack) where T : class
+		{
+			return ReferenceTracker<T>.Track(getInstanceToTrack);
+		}
+	}
+
+	/// <summary>
+	/// A test helper encapsulating correct reference tracking.
+	/// </summary>
+	public struct ReferenceTracker<T> where T : class
+	{
+		private readonly WeakReference weakReference;
+
+		private ReferenceTracker(WeakReference weakReference)
+		{
+			this.weakReference = weakReference;
+		}
+
+		/// <summary>
+		/// <para>
+		/// Use a lambda method to evaluate an expression returning an object instance to track.
+		/// </para>
+		/// <para>
+		/// (If the expression is evaluated directly inside the test method, the runtime will keep it alive
+		/// until the test method returns if the runtime is in debug mode. ReSharper’s test runner runs in
+		/// debug mode when the active solution configuration is Debug.)
+		/// </para>
+		/// </summary>
+		public static ReferenceTracker<T> Track(Func<T> getInstanceToTrack)
+		{
+			if (getInstanceToTrack == null)
+				throw new ArgumentNullException(nameof(getInstanceToTrack));
+
+			return new ReferenceTracker<T>(new WeakReference(getInstanceToTrack.Invoke()));
+		}
+
+		/// <summary>
+		/// Calls <see cref="GC.Collect()"/> and asserts that the tracked instance is still alive.
+		/// </summary>
+		public void AssertStillReferenced()
+		{
+			GC.Collect();
+			if (!weakReference.IsAlive)
+				Assert.Fail("The tracked instance is not longer referenced.");
+		}
+
+		/// <summary>
+		/// Calls <see cref="GC.Collect()"/> and asserts that the tracked instance is no longer alive.
+		/// </summary>
+		public void AssertNoLongerReferenced()
+		{
+			GC.Collect();
+			if (weakReference.IsAlive)
+				Assert.Fail("The tracked instance is still referenced.");
+		}
+
+		/// <summary>
+		/// <para>
+		/// Calls <see cref="GC.Collect()"/> and asserts that the tracked instance is still alive and
+		/// passes it to the specified action.
+		/// </para>
+		/// <para>
+		/// Be careful not to let the tracked instance become reachable via a local variable in the test method
+		/// or it will be kept alive until the end of the test method if the runtime is in debug mode.
+		/// (See <see cref="Track"/>.)
+		/// </para>
+		/// </summary>
+		public void AssertStillReferencedAndDo(Action<T> action)
+		{
+			if (action == null)
+				throw new ArgumentNullException(nameof(action));
+
+			GC.Collect();
+			var target = weakReference.Target;
+			if (target is null)
+				Assert.Fail("The tracked instance is not longer referenced.");
+
+			action.Invoke((T)target);
+		}
+
+		/// <summary>
+		/// <para>
+		/// Calls <see cref="GC.Collect()"/> and asserts that the tracked instance is still alive,
+		/// passes it to the specified function, and returns the value returned by the specified function.
+		/// </para>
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -2,33 +2,353 @@
...
 *.bak
-*.sln.DotSettings.user
-*.DotSettings.user
-
-# Buildscripts artifacts
-buildscripts/**/obj/
-buildscripts/**/bin/
 
 # Tools artifacts
-tools/**/obj/
```
jnm2: All the removed lines were redundant after adding the standard gitignores below.

src/Castle.Facilities.WcfIntegration.Tests/Castle.Facilities.WcfIntegration.Tests.csproj
```diff
@@ -24,6 +24,7 @@
 		<ProjectReference Include="..\Castle.Facilities.Logging\Castle.Facilities.Logging.csproj" />
 		<ProjectReference Include="..\Castle.Facilities.WcfIntegration.Demo\Castle.Facilities.WcfIntegration.Demo.csproj" />
 		<ProjectReference Include="..\Castle.Facilities.WcfIntegration\Castle.Facilities.WcfIntegration.csproj" />
+		<ProjectReference Include="..\Castle.Windsor.Tests\Castle.Windsor.Tests.csproj" />
```
jnm2: Needed in order to use `ReferenceTracker`.

src/Castle.Windsor.Tests/ReferenceTracker.cs
```diff
@@ -0,0 +1,137 @@
...
+		public static ReferenceTracker<T> Track<T>(Func<T> getInstanceToTrack) where T : class
+		{
+			return ReferenceTracker<T>.Track(getInstanceToTrack);
+		}
+	}
+
+	/// <summary>
+	/// A test helper encapsulating correct reference tracking.
+	/// </summary>
+	public struct ReferenceTracker<T> where T : class
```
jnm2: Would be `readonly struct` if you were interested in setting `<LangVersion>latest</LangVersion>` or `<LangVersion>7.2</LangVersion>` in `Directory.Build.props`.

.gitignore
stakx: I think it would be better to keep the `.gitignore` overhaul out of this PR so it stays focused on the actual problem you want to fix.

(Also, in my opinion, after adding all the stock entries below there will be *more* superfluous lines, not fewer. For example, we know that Windsor isn't a ClickOnce project, nor has any C++ artifacts, so the lines for those aren't needed.)

src/Castle.Windsor.Tests/ReferenceTracker.cs
```diff
@@ -0,0 +1,137 @@
...
+
+	public static class ReferenceTracker
+	{
+		/// <summary>
+		/// <para>
+		/// Use a lambda method to evaluate an expression returning an object instance to track.
+		/// </para>
+		/// <para>
+		/// (If the expression is evaluated directly inside the test method, the runtime will keep it alive
+		/// until the test method returns if the runtime is in debug mode. ReSharper’s test runner runs in
```
stakx: (This isn't so much a request to rephrase anything as it's just me being curious, so feel free to ignore.)

I was wondering about the accuracy of _"if the runtime is in debug mode"_. Is it actually the JIT (controlled via `[assembly: Debuggable]`, IIRC), or is it the C# compiler outputting different code, that causes the difference in behavior?

src/Castle.Windsor.Tests/ReferenceTracker.cs
stakx: I am generally in favor of updating to the latest language version, but I happily defer that decision to @jonorossi.

One thing to look out for here is Mono compatibility. There has been a discussion around updating the Windsor build process and possibly set up a build on Mono (we already do this for Core). Mono now ships with the Roslyn compilers, but not always the latest ones. For example, at this time of writing, `<LangVersion>7.3</LangVersion>` wouldn't yet be supported, so updating the language version too far would be a roadblock for setting up a Mono build.

How about holding back this change for now, but open a new issue about converting *all* eligible structs to `readonly` once the project & build setup is definitely ready for it?

src/Castle.Windsor.Tests/ReferenceTracker.cs
```diff
@@ -0,0 +1,137 @@
...
+		/// debug mode when the solution is configuration is Debug.)
```
stakx: `when the solution is configuration is Debug` (2x `is`) &rarr; `when the solution's active configuration is Debug` or perhaps even just `when the Debug configuration is active`

src/Castle.Windsor.Tests/ReferenceTracker.cs
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		/// <summary>
+		/// <para>
+		/// Use a lambda method to evaluate an expression returning an object instance to track.
+		/// </para>
+		/// <para>
+		/// (If the expression is evaluated directly inside the test method, the runtime will keep it alive
+		/// until the test method returns if the runtime is in debug mode. ReSharper’s test runner runs in
+		/// debug mode when the solution is configuration is Debug.)
```
stakx: Same typo as mentioned in review comment above (2x `is`).

src/Castle.Windsor.Tests/ReferenceTracker.cs
```diff
@@ -0,0 +1,137 @@
...
+		/// Calls <see cref="GC.Collect()"/> and asserts that the tracked instance is still alive and
+		/// passes it to the specified action.
+		/// </para>
+		/// <para>
+		/// Be careful not to let the tracked instance become reachable via a local variable in the test method
+		/// or it will be kept alive until the end of the test method if the runtime is in debug mode.
+		/// (See <see cref="Track"/>.)
+		/// </para>
+		/// </summary>
+		public void AssertStillReferencedAndDo(Action<T> action)
```
stakx: When I first saw a use of this method, I didn't quite understand why one would need a combination of an `Assert` + a side effect. I guess it's mostly for clean-up. If so, perhaps it would be more self-explanatory to have a separate `CleanUp` method or similar?, even if that results in two lines (instead of one) at the call sites.

.gitignore
jnm2: Happy to do so. 

The list is a standard and handling it this way makes for easy updates over time as VS and common tools change. I don't think I've ever seen a project remove parts of it. The virtue of cleaning up the manual section is that everything should just work and we're can focus on what is unique to this project. 

Would you prefer I drop the commit?

stakx: Not necessarily drop (I personally don't mind a standard `.gitignore` for the reason you gave, but this is for @jonorossi to decide), but if it is not too much work, submit it as a separate PR of its own.

jnm2: Yes, very happy to do that. https://github.com/castleproject/Windsor/pull/441

src/Castle.Windsor.Tests/ReferenceTracker.cs
jnm2: It is both. The C# compiler performs IL optimization in release mode and not in debug mode, and the JIT performs optimizations unless `Debuggable` disables them.

With VSTest we have an example of running unoptimized IL in a process that has JIT optimizations enabled. This means that it's either all about the JIT operating in debug mode, or else it's a combination of the C# compiler generating unoptimized IL *and* the JIT operating in debug mode.

src/Castle.Windsor.Tests/ReferenceTracker.cs
jnm2: @stakx This method is essential for the tests which need to act on the reference at a later point in time. The whole point of ReferenceTracker is to avoid evaluating that reference directly in the test method.

As to why it also asserts, it seemed like the obvious thing to do if the reference was no longer kept alive. Other options would be to no-op or to return false, but that just complicates the implementations of the tests.

Yet another alternative which I didn't do because it seemed complex was to extend this helper API to have some sort of keep-alive methods.

src/Castle.Windsor.Tests/ReferenceTracker.cs
jnm2: Done, see https://github.com/castleproject/Windsor/issues/442.

jnm2:
@stakx Updated. Sorry, I think I got all the whitespace changes. It's an uphill battle against my tooling. :'( If you know of a faster way to undo whitespace trimming than `git reset -p`, let me know!

src/Castle.Windsor.Tests/ReferenceTracker.cs
stakx: Interesting! Thanks for checking up on this & letting me know. Quite happy to leave the comment as is.

src/Castle.Windsor.Tests/ReferenceTracker.cs
stakx: If it's not trivial to split up `AssertStillReferencedAndDo` into an `AssertStillReferenced` and `CleanUp` (or whatever it would be called), then please ignore my suggestion and leave the method as is. It probably doesn't make any sense to bloat this helper class' API needlessly.

stakx:
@jnm2 - With repos having a history of bad indentation / whitespace, I usually disable any "trim trailing whitespace" setting etc. so whitespace won't get silently changed behind my back when saving files. Sometimes, I still cause whitespace changes by actual line edits. Those, I usually just undo by hand, as they're usually only few in number.

Perhaps someday soon we will fix whitespace / indentation issues across the whole solution to match any `.editorconfig` file we might have by then, and from that glorious day onward tooling should do what it's supposed to do. :-)

(Please apologize if I appear overly pedantic, I know whitespace isn't *that* important in the grand scheme of things, but it's nice to have clean diffs without minimal noise.)

src/Castle.Windsor.Tests/ReferenceTracker.cs
jnm2: Yeah; the specific issue with splitting it into two methods is that it's no longer atomic; there will then be two separate calls to WeakReference.Target and the target may be alive for the assert check and dead for the `Do`/`CleanUp` work. To split it up, we'd need to add KeepAlive-style methods. maybe like:

```c#
var target = ReferenceTracker.Track(() => Container.Resolve<ThingWhichMustNotBeNamed>());

using (target.AssertStillReferencedAndStartKeepingAlive())
{
    target.Do(thing => container.Release(thing));
} // Stop keeping it alive

target.AssertNoLongerReferenced();
```

Oops, now we have `AssertStillReferencedAndStartKeepingAlive` instead of `AssertStillReferencedAndDo`. I didn't think that far ahead, but we can't avoid making a decision about what to do atomically if the thing is no longer alive when the test expects it to be, and asserting seemed like it put the least burden on the test method.

jnm2:
No problem at all! I understand. I'm similarly pedantic, only in the direction of trimming rather than of preventing whitespace changes. If formatting (whitespace or otherwise) is really bad, I'll put it all in its own commit for the whole PR.

PR #441, opened 2018-10-11
Update .gitignore so that NCrunch .user files are ignored

jnm2:
NCrunch kept adding a file tracked by git, so I added this standard list. Handling it this way makes for easy updates over time as VS and common tools change.

The virtue of cleaning up the manual section is that everything should just work and we can focus on what is unique to this project.

/cc @jonorossi, @stakx

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,12 +4,11 @@ AssemblyInfo.cs
 
 # Standard VS.NET and ReSharper Foo
 .vs/*
+*.user
 src/*/obj
 src/*/bin
 samples/*/obj
 samples/*/bin
-*.csproj.user
-*ReSharper.user
 _ReSharper*
 *.suo
 *.cache
@@ -18,8 +17,6 @@ _ReSharper*
 # Leftovers from merging
 *.orig
 *.bak
-*.sln.DotSettings.user
-*.DotSettings.user
 
 # Buildscripts artifacts
 buildscripts/**/obj/
--- a/.gitignore
+++ b/.gitignore
@@ -1,31 +1,10 @@
 
 /build
-AssemblyInfo.cs
 
 # Standard VS.NET and ReSharper Foo
 .vs/*
 *.user
-src/*/obj
-src/*/bin
-samples/*/obj
-samples/*/bin
+bin/
+obj/
 _ReSharper*
 *.suo
-*.cache
-* Thumbs.db
-
-# Leftovers from merging
-*.orig
-*.bak
-
-# Buildscripts artifacts
-buildscripts/**/obj/
-buildscripts/**/bin/
-
-# Tools artifacts
-tools/**/obj/
-tools/**/build/
-
-# Docs artifacts
-docs/obj/
-docs/bin/


jonorossi:
> NCrunch kept adding a file tracked by git

Really, NCrunch calls git add on the files it writes?

> so I added this standard list. Handling it this way makes for easy updates over time as VS and common tools change.
> 
> The virtue of cleaning up the manual section is that everything should just work and we can focus on what is unique to this project.

I've always disliked these "standard" gitignores, I feel it legitimises tools writing crap all over your source directory. A 350 line gitignore not only needs to be parsed and compared by git on every operation it does, but it contains cruft like `project.lock.json` which is dead along with heaps of other tools.

I do however think the gitignore in the repo at the moment should be simplified a bit.

jnm2:
> I feel it legitimises tools writing crap all over your source directory.

I'm not trying to change your mind, but I am not sure I follow. What should tools do instead?

stakx:
> What should tools do instead?

Most tools could probably make life easier for `.gitignore` minimalists by putting their artifacts either under `obj/` or in a dot directory (such as Git does with `.git/`, or VS Code with `.vscode/`). One would only need two `.gitignore` rules at most, that way.

But the real question here, of course, isn't what tools should do (that's out of our control), but whether we should cater for our contributors' choice of tools and therefore have a kitchen sink `.gitignore` in advance, or whether they should adapt to *our* choice of `.gitignore` and pay some additional care to staging their commits properly.

The solution will probably be somewhere between these two extremes. Granted, a stock `.gitignore` means we likely won't have to deal with it again for some time. I personally start disliking `.gitignore` files when they get so large that you can no longer see the forest for the trees.

So, why not start with a stock `.gitignore`, then go towards a more minimal file by eliminating everything that *clearly* won't ever apply to this particular project (such as rules for C++ build artifacts)?

stakx:
P.S.: Once we have decided on a `.gitignore`, why not put a copy of it in castleproject/Home so other Castle projects can easily find & reuse it?

jonorossi:
> The emphasis is that next time you can update from gitignore.io without having to think about which parts you had originally deleted from the past standard gitignore.io. I used to maintain them manually, but it's a lot of busy work that's not for me. If you prefer it, that's totally fine.

There are [424 different gitignore templates available from that site](https://github.com/dvcs/gitignore/tree/master/templates), the "VisualStudio" one is just one of them.

I guess the problem I've seen is that not all repos use a single language and set of tooling so these gitignores just become silly with hundreds or thousands of lines. I have also seen them cause problems where an ignore excluded something that should have been committed, however it does seem like the templates have been changed to comment out things not as clear cut which you have to go in an uncomment those.

> I'm not trying to change your mind, but I am not sure I follow. What should tools do instead?

Generally if I use a tool that writes cache into the source directory (rather than say temp) I'll add it to my user gitignore rather than get every project to add it to their gitignores. I'll generally keep gitignores in a repo contained to artifacts that the build scripts output.

That said, if you feel this is a valuable change to Windsor (which you must or you wouldn't have submitted it) I don't want to get in your way from contributing because I'm not a user of the project anymore and we need some fresh contributors.

I'll make some comments on the changes.

.gitignore
```diff
@@ -2,33 +2,353 @@
 /build
 AssemblyInfo.cs
```
jonorossi: We can remove the AssemblyInfo.cs exclusion now, we used to generate these with the old tooling.

.gitignore
```diff
@@ -2,33 +2,353 @@
...
-*.sln.DotSettings.user
-*.DotSettings.user
-
-# Buildscripts artifacts
-buildscripts/**/obj/
-buildscripts/**/bin/
 
 # Tools artifacts
-tools/**/obj/
 tools/**/build/
```
jonorossi: I haven't had a look but what is in `tools/**/build`? I'm going to guess the project was modified to write to build from the standard bin directory?

.gitignore
```diff
@@ -2,33 +2,353 @@
...
-src/*/obj
-src/*/bin
-samples/*/obj
-samples/*/bin
-*.csproj.user
-*ReSharper.user
-_ReSharper*
-*.suo
 *.cache
 * Thumbs.db
```
jonorossi: Thumbs.db is an example of something I'd handle on a system globally, we don't even have it in Core.

.gitignore
```diff
@@ -2,33 +2,353 @@
...
-.vs/*
-src/*/obj
-src/*/bin
-samples/*/obj
-samples/*/bin
-*.csproj.user
-*ReSharper.user
-_ReSharper*
-*.suo
 *.cache
```
jonorossi: Not sure what writes a .cache file, do you know?

.gitignore
```diff
@@ -2,33 +2,353 @@
...
 * Thumbs.db
 
 # Leftovers from merging
 *.orig
 *.bak
```
jonorossi: We can probably get rid of these 2 as well. I've never seen a merge leave these behind unless you abort it wrong.

.gitignore
```diff
@@ -2,33 +2,353 @@
 /build
```
jonorossi: It may be that we'll just be left with one of our own entries.

jnm2:
@jonorossi 

> That said, if you feel this is a valuable change to Windsor (which you must or you wouldn't have submitted it) I don't want to get in your way from contributing because I'm not a user of the project anymore and we need some fresh contributors.

I did, but I hadn't had a reason to question that this was the obvious thing to do. It's an interesting idea that if you use a tool, you should add it to your user gitignore. I think for projects I manage I'd want to add gitignores for tools users are *likely* to be using in my particular repo, but I can see it both ways.

Why don't I remove the stock set and instead just add  the pattern `*.user`? That's a standard way (in the VS ecosystem) to mark files that hold settings that are not intended to be shared in source control.

jonorossi:
> I think for projects I manage I'd want to add gitignores for tools users are _likely_ to be using in my particular repo, but I can see it both ways.

That is the pragmatic option I've gone with recently, which is why I'll include `.vs` and still include `*.user` and `*.suo`. ReSharper has also been a very heavily used tool with Castle over the last 10+ years so it always made the cut even though from a purest viewpoint it doesn't belong.

> Why don't I remove the stock set and instead just add the pattern `*.user`? That's a standard way (in the VS ecosystem) to mark files that hold settings that are not intended to be shared in source control.

Sounds good, you'll see Castle Core has it that way. I think it would be good if Windsor's was more like Castle Core's.

.gitignore
jnm2: Visual Studio, but it's always been inside the obj folder as far as I know, so this looks redundant.

jonorossi:
@jnm2 just checking in on this pull request. Had you finished? I was hoping to see all those separate bin/obj rules disappear to be just like Castle Core's gitignore.

jnm2:
`/bin` and `/obj` were patterns that came from the stock gitignore. Did you want me to add those entries besides adding `*.user`?

jnm2:
Done.

jonorossi:
Cheers

PR #450, opened 2018-10-26
Fixed first-chance HandlerException for optional parameters

### LINKED ISSUE #449: First-chance HandlerException thrown for optional parameters
It's good practice to minimize exception throwing and handling where possible. First-chance exceptions cause the debugger to break when you're debugging some unrelated issue which requires you to turn off Just My Code. In order to get to the exception or breakpoint you care about, you have to skip through many exceptions you don't care about. It can make an otherwise short series of debugging sessions frustrating, especially if the debugger is a bit sluggish.

This is a regular thorn in the side caused by other libraries as well, including the BCL itself. See `https://github.com/dotnet/corefx/issues/19928#issuecomment-330630112` and lower, for instance.

Culprit:

https://github.com/castleproject/Windsor/blob/09751a40229cc9867e4a16f8c3a23cd92ec87adb/src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs#L427-L435

(The `Try` name implies that it is not an exceptional case if the handler cannot be gotten from the kernel, and therefore an exception should not be thrown. Optional parameters are indeed not an exceptional case, so it would be better to follow the `Try` pattern here instead of throwing.)

Repro code that works outside the debugger, suitable for unit testing:

```c#
using System;
using System.Collections.Generic;
using System.Linq;
using System.Runtime.ExceptionServices;
using Castle.Windsor;

public static class Program
{
    public sealed class SomeService
    {
        public SomeService(IEqualityComparer<string> optionalParam = null)
        {
        }
    }

    public static void Main()
    {
        using (var container = new WindsorContainer())
        {
            container.Register(Component.For<SomeService>());

            var service = AssertNoFirstChanceExceptions(() => container.Resolve<SomeService>());

            container.Release(service);
        }
    }

    private static T AssertNoFirstChanceExceptions<T>(Func<T> function)
    {
        T returnValue;
        var firstChanceExceptions = new List<Exception>();

        var handler = new EventHandler<FirstChanceExceptionEventArgs>((sender, e) =>
            firstChanceExceptions.Add(e.Exception));

        AppDomain.CurrentDomain.FirstChanceException += handler;
        try
        {
            returnValue = function.Invoke();
        }
        finally
        {
            AppDomain.CurrentDomain.FirstChanceException -= handler;
        }

        if (firstChanceExceptions.Any())
        {
            Assert.Fail("First-chance exceptions thrown:\r\n\r\n"
                + string.Join("\r\n\r\n", firstChanceExceptions));
        }

        return returnValue;
    }
}
```

jnm2:
Fixes #449.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Castle Windsor Changelog
 
+## Unreleased
+
+Bugfixes:
+- Fixed first-chance HandlerException for optional parameters (@jnm2, #450)
+
 ## 5.0.0-beta001 (2018-10-26)
 
 Enhancements:
--- a/src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
+++ b/src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
@@ -361,12 +361,7 @@ namespace Castle.MicroKernel.Resolvers
 
 		private object ResolveFromKernelByType(CreationContext context, ComponentModel model, DependencyModel dependency)
 		{
-			IHandler handler;
-			try
-			{
-				handler = TryGetHandlerFromKernel(dependency, context);
-			}
-			catch (HandlerException exception)
+			if (!TryGetHandlerFromKernel(dependency, context, out var handler))
 			{
 				if (dependency.HasDefaultValue)
 				{
@@ -377,8 +372,7 @@ namespace Castle.MicroKernel.Resolvers
 						"Missing dependency.{2}Component {0} has a dependency on {1}, which could not be resolved.{2}Make sure the dependency is correctly registered in the container as a service, or provided as inline argument.",
 						model.Name,
 						dependency.TargetItemType,
-						Environment.NewLine),
-					exception);
+						Environment.NewLine));
 			}
 
 			if (handler == null)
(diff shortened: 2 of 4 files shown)


src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
```diff
@@ -361,12 +361,7 @@ private object ResolveFromKernelByName(CreationContext context, ComponentModel m
 
 		private object ResolveFromKernelByType(CreationContext context, ComponentModel model, DependencyModel dependency)
 		{
-			IHandler handler;
-			try
-			{
-				handler = TryGetHandlerFromKernel(dependency, context);
-			}
-			catch (HandlerException exception)
+			if (!TryGetHandlerFromKernel(dependency, context, out var handler))
```
jnm2: The `return false` case looks almost identical to the `handler = null; return true` case below it.
There's a potentially a better factoring to indicate which of the two messages should be used?

jnm2: ⚠ Oh, this may not be a valid refactoring. I forgot to consider HandlerExceptions that may be thrown by methods called by `TryGetHandlerFromKernel`. We might need to catch `HandlerException` within `TryGetHandlerFromKernel` and return false;

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
```diff
@@ -424,23 +418,37 @@ private object ResolveFromParameter(CreationContext context, ComponentModel mode
...
 
 			// make a best effort to find another one that fit
 
-			var handlers = kernel.GetHandlers(dependency.TargetItemType);
+			IHandler[] handlers;
+			try
+			{
+				handlers = kernel.GetHandlers(dependency.TargetItemType);
+			}
+			catch (HandlerException)
```
jnm2: To make this a valid refactoring, we have to be sure that this method can't throw `HandlerException` indirectly.
If you know that `kernel.GetHandlers` would not throw `HandlerException`, we can remove the `try`. Same for the `try` above.

jonorossi: `DefaultNamingSubSystem` doesn't appear to, however a custom `INamingSubSystem` or `IHandlersFilter` may. Sounds like a sensible way to write it.

jnm2:
🎉 Updated!

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@ Breaking Changes:
   - Changed `IComponentModelBuilder.BuildModel` to use `Arguments` instead of `IDictionary`
   - Changed `ILazyComponentLoader.Load` to use `Arguments` instead of `IDictionary`
 
+Bugfixes:
+- Fixed first-chance HandlerException for optional parameters (@jnm2, #450)
```
jonorossi: @jnm2 You've added this inside the 5.0.0-beta001 section which has been released.

PR #462, opened 2019-02-05
Upgrade ASP.NET Core facility to require ASP.NET Core 2.2

jonnii:
This updates the aspnet core facility to point to 2.2.


jonnii:
I'm not sure if this is the best solution but the pinned versions in the csproj are making it impossible to use this package without automatic downgrades against an aspnet core 2.2 project.

jonnii:
@jonorossi i got this to work, I was doing something stupid. Any idea when this will all be released officially?

jonorossi:
> Any idea when this will all be released officially?

Just logged #464 to track that, decided to just wait until someone asked for it.

PR #480, opened 2019-03-28
Finding the controller should be made in a case insensitive way

Yitzchok:
This is for AspNetMvcFacility.
When searching for a controller matching the url from the web request we shouldn't care about the case.

### DIFF
--- a/src/Castle.Facilities.AspNet.Mvc/AspNetMvcControllerFactory.cs
+++ b/src/Castle.Facilities.AspNet.Mvc/AspNetMvcControllerFactory.cs
@@ -24,7 +24,6 @@ namespace Castle.Facilities.AspNet.Mvc
 	using System.Web.Routing;
 	using System.Web.SessionState;
 
-	using Castle.Core.Internal;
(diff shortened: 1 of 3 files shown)


src/Castle.Facilities.AspNet.Mvc/AspNetMvcControllerFactory.cs
```diff
@@ -100,7 +101,7 @@ private Type FindControllerType(string controllerName)
 		{
 			foreach (var type in this.controllerCandidateTypes)
 			{
-				if (type.Name == controllerName)
+				if (type.Name.EqualsText(controllerName))
```
jonorossi: I'd just use `type.Name.Equals(controllerName, StringComparison.OrdinalIgnoreCase)`. That extension method seems a little unnecessary to even exist, and shouldn't be public.

Yitzchok: Fixed

Yitzchok:
I added it to the changelog

Issue #487, opened 2019-05-07
Remove the dependency on ASPNET Core  in Asp.NET Core Facility, just use .NET Core


schwarzie2478:
I took some initiative to add the facility to the list of supported third party containers in the repository https://github.com/aspnet/Extensions while I was investigating how to use it for my own work.

I got a question from david fowler about the current implementation.   
I'm not really the guy to answer him,  so I leave this here to let you guys  decide.

Can the code be changed to not depend on  asp.net core and only on the .NET core dependency injection infrastructure?

Pull request with the change if somebody is interested: [https://github.com/aspnet/Extensions/pull/1632](https://github.com/aspnet/Extensions/pull/1632)

jonorossi:
There is no active maintainer for this facility, there unfortunately is no "you guys".

Your pull request obviously makes breaking changes, if you want to see this go in there needs to be a plan for this facility. It sounds like the facility won't work with ASP.NET Core 3.0 (#493), if the replacement facility is a `Microsoft.Extensions.DependencyInjection` integration then how that integration will work should be described.

jonorossi:
Closing as `Castle.Windsor.Extensions.DependencyInjection` is the path forward for ASP.NET Core 3.x.

PR #488, opened 2019-05-19
Separated the .NET Core handling from the specific ASP.NET additional…

schwarzie2478:
… configuration options.
See issue: castleproject/Windsor#487

As a start I basically separated the functionality into two facilities: NetCoreFacility & AspNetCoreFacitlity.

I first renamed the original project to NetCore, threw out all AspeNetCore dependent code.
Then I added the original project again and removed the double definitions.
I had to make some properties more accesible to make sure all code still worked.

So if you want to use Windsor in an non-web application, you no longer are required to include AspNetCore packages.

More work is needed to further clean up and separte the documentation.

Could someone review this already perhaps. I'll try to continue to send further updates.

### DIFF
--- a/Castle.Windsor.sln
+++ b/Castle.Windsor.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2024
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28803.452
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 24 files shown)


schwarzie2478:
I've made the split as logical as possible without trying to change too much of the original structure of the projects.

So do we need a separate test project for testing the .NET core facility?
There is still some code in the NETCore project that will only hold ASP.Net Core things, but it doesn't directly reference   those assemblies.  Do we move this code?  

Now I have made a .NET Core facility and one ASP.NET Core deriving from it. Do we want this, or remove the ASPCore one and just leave the asp.net Core specific extentions methods and such in that project.

jonorossi:
As I said in #487 we really need a plan before throwing more code at the problem. There appears to be quite a few people using Windsor and ASP.NET Core now that I think a discussion would be valuable, would be great to see a single DependencyInjection facility.

I'm going to close this for now, this also now has conflicts.

PR #505, opened 2019-11-08
DefaultComponentActivator sets nonpublic props too

ivan-danilov:
The reason is that it is the model's decision what properties should be injected, not activator's. By default, model excludes non-public properties. But if some customization was applied that requested injecting through non-public setter - now it requires providing a custom activator, which is not trivial to write and use (e.g. you need to select the activator for your components correctly).


jonorossi:
Sorry for not replying before now, since this pull request I'm going to close it, but feel free to reopen it. We need some unit tests for these changes.

PR #510, opened 2019-12-10
TypedFactory works with multi-Dispose calls and Release after Dispose

### LINKED ISSUE #458: Typed factory should allow disposing multiple times
According to .NET Guidelines a dispose method should tolerate being called multiple times without throwing an exception.  
This exception should not be thrown when `method` equals `FactoryMethod.Dispose`:

https://github.com/castleproject/Windsor/blob/603d78d491790c88cff604051c92115e8bb5d9f9/src/Castle.Windsor/Facilities/TypedFactory/Internal/TypedFactoryInterceptor.cs#L55

ivan-danilov:
Fixes #458


ivan-danilov:
It does not seem that the fail has anything to do with the changes in this PR...

PR #522, opened 2020-05-06
Treat null as a valid Nullable<T> dependency value

### LINKED ISSUE #521: Null should be treated as a valid dependency value when passed to a Nullable<T> parameter
Failing test:

```cs
[Test]
public void Null_may_be_specified_for_non_optional_System_Nullable_typed_factory_parameter()
{
	Container.Register(
		Component.For<DependencyFromContainer>(),
		Component.For<ComponentWithNonOptionalNullableParameter>(),
		Component.For<ComponentWithNonOptionalNullableParameter.Factory>().AsFactory());

	var factory = Container.Resolve<ComponentWithNonOptionalNullableParameter.Factory>();
	factory.Invoke(nonOptionalNullableParameter: null);
}

public sealed class DependencyFromContainer
{
}

public sealed class ComponentWithNonOptionalNullableParameter
{
	public delegate ComponentWithNonOptionalNullableParameter Factory(int? nonOptionalNullableParameter);

	public ComponentWithNonOptionalNullableParameter(int? nonOptionalNullableParameter, DependencyFromContainer dependencyFromContainer)
	{
	}
}
```

jnm2:
Fixes #521

Not sure if this is the best approach. Directions welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-...
+Enhancements:
+- Enable explicitly specified null values to satisfy System.Nullable<> dependencies (@jnm2, #521)
 
 ## 5.0.1 (2019-09-18)
 
--- a/src/Castle.Windsor.Tests/SubResolverTestCase.cs
+++ b/src/Castle.Windsor.Tests/SubResolverTestCase.cs
@@ -45,30 +45,68 @@ namespace Castle.MicroKernel.Tests
 
 			Assert.AreEqual(HandlerState.Valid, handler.CurrentState);
 		}
-	}
 
-	public class Foo
-	{
-		private int bar;
-
-		public Foo(int bar)
+		public class Foo
 		{
-			this.bar = bar;
-		}
-	}
+			private int bar;
 
-	public class FooBarResolver : ISubDependencyResolver
-	{
-		public int? Result;
-
-		public bool CanResolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
-		{
-			return Result != null;
+			public Foo(int bar)
+			{
+				this.bar = bar;
+			}
 		}
 
-		public object Resolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
+		public class FooBarResolver : ISubDependencyResolver
 		{
-			return Result.Value;
+			public int? Result;
+
+			public bool CanResolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
+			{
+				return Result != null;
+			}
+
+			public object Resolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
+			{
+				return Result.Value;
+			}
+		}
+
+		[Test]
+		public void Sub_resolver_can_provide_null_as_the_value_to_use()
+		{
+			IKernel kernel = new DefaultKernel();
+			kernel.Resolver.AddSubResolver(new NullResolver());
+
+			kernel.Register(Component.For<ComponentWithDependencyNotInContainer>());
+
+			Assert.Null(kernel.Resolve<ComponentWithDependencyNotInContainer>().DependencyNotInContainer);
+		}
+
+		public sealed class ComponentWithDependencyNotInContainer
+		{
+			public ComponentWithDependencyNotInContainer(DependencyNotInContainer dependencyNotInContainer)
+			{
+				DependencyNotInContainer = dependencyNotInContainer;
+			}
+
+			public DependencyNotInContainer DependencyNotInContainer { get; }
+		}
+
+		public sealed class DependencyNotInContainer
+		{
+		}
+
+		private sealed class NullResolver : ISubDependencyResolver
+		{
+			public bool CanResolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
+			{
+				return true;
+			}
+
+			public object Resolve(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
+			{
+				return null;
+			}
 		}
 	}
 }
--- a/src/Castle.Windsor.Tests/TestUtils.cs
+++ b/src/Castle.Windsor.Tests/TestUtils.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2018 Castle Project – http://www.castleproject.org/
+﻿// Copyright 2018–2020 Castle Project – http://www.castleproject.org/
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !NETCOREAPP1_0 // FirstChanceException event was added in .NET Core 2.0
 namespace CastleTests
 {
 	using System;
@@ -20,11 +19,13 @@ namespace CastleTests
 	using System.Linq;
 	using System.Runtime.ExceptionServices;
 	using System.Text;
+	using System.Text.RegularExpressions;
 
 	using NUnit.Framework;
 
 	public static class TestUtils
 	{
+#if !NETCOREAPP1_0 // FirstChanceException event was added in .NET Core 2.0
 		public static void AssertNoFirstChanceExceptions(Action action)
 		{
 			var firstChanceExceptions = new List<Exception>();
@@ -57,6 +58,11 @@ namespace CastleTests
 				Assert.Fail(message.ToString());
 			}
 		}
+#endif
+
+		public static string ConvertToEnvironmentLineEndings(this string value)
+		{
+			return Regex.Replace(value, @"\r?\n", Environment.NewLine);
+		}
 	}
 }
-#endif
--- a/src/Castle.Windsor/Core/Internal/TypeUtil.cs
+++ b/src/Castle.Windsor/Core/Internal/TypeUtil.cs
@@ -22,6 +22,19 @@ namespace Castle.Core.Internal
 	public static class TypeUtil
 	{
 		/// <summary>
+		/// Indicates whether the type is explicitly marked as nullable. Currently only detects nullable value types via
+		/// <see cref="Nullable{T}"/>, but could be expanded to detect C# nullable reference types.
+		/// </summary>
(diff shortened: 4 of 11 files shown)


src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
```diff
@@ -321,7 +321,11 @@ private bool CanConvertParameter(Type type)
 		private bool CanResolve(DependencyModel dependency, object inlineArgument)
 		{
 			var type = dependency.TargetItemType;
-			if (inlineArgument == null || type == null)
+			if (inlineArgument == null)
+			{
+				return type.IsNullable();
```
jonorossi, line 326: Could `type` be null here?

src/Castle.Windsor/Core/Internal/TypeUtil.cs
```diff
@@ -22,7 +22,20 @@ namespace Castle.Core.Internal
...
+		/// Indicates whether the type is explicitly marked as nullable. Currently only detects nullable value types via
+		/// <see cref="Nullable{T}"/>, but could be expanded to detect C# nullable reference types.
+		/// </summary>
+		internal static bool IsNullable(this Type type)
+		{
+			// For now, only detects nullable value types. Detecting nullable reference types is complicated.
+			// Spec: https://github.com/dotnet/roslyn/blob/version-3.2.0/docs/features/nullable-metadata.md
+			// Reflection API proposal: https://github.com/dotnet/runtime/issues/29723
+
+			return type.GetTypeInfo().IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>);
```
jonorossi, line 34: Any reason not to just use `Nullable.GetUnderlyingType(type) != null` inline?

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
```diff
@@ -301,25 +300,28 @@ private bool HasComponentInValidState(string key, DependencyModel dependency, Cr
 			return IsHandlerInValidState(handler) && handler.IsBeingResolvedInContext(context) == false;
 		}
 
-		private object ResolveCore(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency)
+		private bool TryResolveCore(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency, out object value)
```
jonorossi, line 303: This appears to have no functional difference, however is this a good change? Are you sure all of the methods that this calls don't throw exceptions?

src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
jonorossi, line 326: This change appears fine, I can't think of anything this would cause a problem for. Since it is in a core area used by everything could you please add a unit test for the core functionality in addition to your typed factory one.

src/Castle.Windsor/Core/Internal/TypeUtil.cs
jnm2: I guess this is a premature optimization (avoids accessing the generic type arguments array). Castle.Core implements `IsNullableType()` this way though.

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
jnm2: This change is required in order to distinguish between returning a null value as success and returning a null value as failure. The change is not meant to have any implications with regard to exception handling.

src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
jnm2: Good find, thanks!

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
jnm2: Would you prefer I return `(bool Success, object Value)` or introduce an `Option` type into the codebase? (`Option.None<object>()` for failure to resolve versus `Option.Some<object>(null)` for successful resolution of the value `null`)

src/Castle.Windsor/Core/Internal/TypeUtil.cs
jnm2: Ah, as far as the inline part. I was hoping that one day, nullable reference types would also be detected by this extension method.

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
jonorossi: > This change is required in order to distinguish between returning a null value as success and returning a null value as failure. It isn't meant to have any implications with regard to exception handling.

Ah yes, that is the important part of your change, missed that. I haven't had a good enough look/think, but do you think this could break the behaviour of the `ISubDependencyResolver` contract because returning null from them now does something different?

> Would you prefer I return ...

I don't mind the well established .NET `bool TryXXX` pattern.

src/Castle.Windsor/Core/Internal/TypeUtil.cs
jonorossi: Haha, looks like I added `IsNullableType` to Castle Core two years ago after refactoring the same reflection check obviously without even a second thought.

I'll leave it up for you to decide, but I just thought you'd then just do it inline without this extra method, because nullable reference types are the same `System.Type` with attributes applied to the `ParameterInfo` so wouldn't be done in this method.

src/Castle.Windsor/MicroKernel/Resolvers/DefaultDependencyResolver.cs
jnm2: I thought it wouldn't break because of the `subResolver.CanResolve` check. Are there cases where `subResolver.CanResolve` returns true and `subResolver.Resolve` returns null today? If so, maybe they are also bugs fixed by this PR?

jnm2: If you'd like me to research this, any ideas on likely container configurations would be great. If not, I could return `value is object` and leave a note that this preserves previous behavior but we haven't investigated whether the previous behavior is correct.

jonorossi: Probably. Could you check the unit tests for sub resolvers and possibly add an extra one if needed to capture this design decision.

src/Castle.Windsor/Core/Internal/TypeUtil.cs
jnm2: I would assume this method would begin requiring the ParameterInfo to be passed at that point and almost wrote this in the comment :D Sadly, the attributes are sometimes not applied to the ParameterInfo but rather to the method or type for compression purposes. But my thought was that passing a second ICustomAttributeProvider parameter would allow the right traversals to take place.

I'll keep this since that's okay with you and I feel like it could increase the chances of nullable reference types being recognized.

src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
jnm2: I haven't been able to figure out how to write a test that created a dependency with a null type, which is what would be required. This code creates a DependencyModel with a null type:

`Component.For(typeof(ClassWithListConstructor)).Named("key")` 

But I'm not sure how to go from there to making the DependencyModel be a dependency of another component.

jnm2: I'll just fix without a test unless you know where to point me.

jnm2:
@jonorossi Updated. There is one unresolved question above. I used `git commit --fixup` to make reviewing easier. Are you okay with me rebasing and force pushing to apply the fixups after approval and before merge?

src/Castle.Windsor.Tests/TestUtils.cs
```diff
@@ -12,19 +12,20 @@
...
 	using System.Linq;
 	using System.Runtime.ExceptionServices;
 	using System.Text;
+	using System.Text.RegularExpressions;
 
 	using NUnit.Framework;
 
 	public static class TestUtils
 	{
+#if !NETCOREAPP1_0 // FirstChanceException event was added in .NET Core 2.0
```
jnm2, line 28: 2.1 is the oldest .NET Core runtime Microsoft still supports. Of course you can decide whether you want to support each version for longer or shorter than Microsoft, but I would recommend adding `netcoreapp2.1` and `netcoreapp3.1` targets to the test project so that the whole suite of tests runs on each supported version of the runtime.

jonorossi: We are going through this with Castle Core right now removing support for unsupported .NET Core versions and adding newer .NET Standard 2.x targets, we'll need to do that here once we are done and released there.

src/Castle.Windsor.Tests/SubResolverTestCase.cs
```diff
@@ -1,3 +1,4 @@
+
```
jonorossi, line 1: Unneeded blank line at beginning of file.

src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
jonorossi: Sounds good, you've covered a good number of scenarios, thanks!

jonorossi:
> I used `git commit --fixup` to make reviewing easier. Are you okay with me rebasing and force pushing to apply the fixups after approval and before merge?

Definitely, I love it when people rebase their commits in a pull request to present a clean history of their work to merge. Feel free to update the changelog too with this change.

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-...
+Enhancements:
+- Enable explicitly specified null values to satisfy System.Nullable<> dependencies (@jnm2, #521)
```
jnm2, line 6: Would you consider this an enhancement, and should I add any more detail?

jonorossi: Enhancement sounds good since this wasn't a supported feature.

That's enough detail. Usually if you think there is more detailed required then I usually push people to add that to the docs rather than the changelog.

jnm2:
@jonorossi Just checking that you're not waiting for me to be doing something.

jonorossi:
@jnm2 apologies for not getting to this before now, everything looks good, merged.

jnm2:
No worries, thank you!

Issue #533, opened 2020-07-01
Extensions.DependencyInjection: "An item with the same key has already been added" exception related to scoped lifestyle


mdavis:
When using the new Extensions.DependencyInjection package with an ASP.NET Core 3.1 web application, if multiple requests come in as soon as the application starts up, I'm seeing ArgumentExceptions with the message "An item with the same key has already been added." specifying some service with a scoped lifestyle that was being resolved.

Here is an example error log:

```
[2020-07-01 14:51:50.7450] (ERROR) Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware: An unhandled exception has occurred while executing the request.
System.ArgumentException: An item with the same key has already been added. Key: NoopClaimsTransformation / IClaimsTransformation
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Castle.MicroKernel.Lifestyle.Scoped.ScopeCache.set_Item(Object id, Burden value)
   at Castle.Windsor.Extensions.DependencyInjection.Scope.ExtensionContainerScope.GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
   at Castle.MicroKernel.Lifestyle.ScopedLifestyleManager.Resolve(CreationContext context, IReleasePolicy releasePolicy)
   at Castle.MicroKernel.Handlers.DefaultHandler.ResolveCore(CreationContext context, Boolean requiresDecommission, Boolean instanceRequired, Burden& burden)
   at Castle.MicroKernel.Handlers.DefaultHandler.Resolve(CreationContext context, Boolean instanceRequired)
   at Castle.MicroKernel.Handlers.AbstractHandler.Resolve(CreationContext context)
   at Castle.MicroKernel.Resolvers.DefaultDependencyResolver.ResolveFromKernelByType(CreationContext context, ComponentModel model, DependencyModel dependency)
   at Castle.MicroKernel.Resolvers.DefaultDependencyResolver.ResolveFromKernel(CreationContext context, ComponentModel model, DependencyModel dependency)
   at Castle.MicroKernel.Resolvers.DefaultDependencyResolver.TryResolveCore(CreationContext context, ISubDependencyResolver contextHandlerResolver, ComponentModel model, DependencyModel dependency, Object& value)
[...]

PR #547, opened 2020-09-08
Add lock to scoped burdens to ensure only one instance is created per scope

generik0:
Add lock to scoped burdens to ensure only one instance is created per scope

#533

### DIFF
--- a/src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
+++ b/src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.2;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.2;netcoreapp3.0;netcoreapp3.1;net461;net471;net48</TargetFrameworks>
 
     <IsPackable>false</IsPackable>
 
--- a/src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
+++ b/src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.2;netcoreapp3.0;netcoreapp3.1;net461;net471;net48</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.2;netcoreapp3.0;netcoreapp3.1;net461</TargetFrameworks>
 
     <IsPackable>false</IsPackable>
 
--- a/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
+++ b/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
@@ -70,8 +70,14 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Scope
 				{
 					burden = scopeCache[model];
 				}
+				else
+				{
+					burden = createInstance((_) => {});
+					scopeCache[burden] = burden;
+				}
 				if (burden != null) return burden;
-				scopeCache[model] = burden = createInstance((_) => {});
+				burden = createInstance((_) => {});
+				scopeCache[model] = burden;
 				return burden;
 			}
 		}
--- a/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
+++ b/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
@@ -65,20 +65,23 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Scope
 		{
 			lock (scopeCache)
 			{
-				Burden burden = null;
-				if(model.Configuration.Attributes.Get(TransientMarker) != bool.TrueString )
+				
+				// Add transient's burden to scope so it gets released
+				if (model.Configuration.Attributes.Get(TransientMarker) == bool.TrueString)
 				{
-					burden = scopeCache[model];
+					var transientBurden = createInstance((_) => {});
+					scopeCache[transientBurden] = transientBurden;
+					return transientBurden;
 				}
-				else
+
+				var scopedBurden = scopeCache[model];
+				if (scopedBurden != null)
 				{
-					burden = createInstance((_) => {});
-					scopeCache[burden] = burden;
+					return scopedBurden;
 				}
-				if (burden != null) return burden;
-				burden = createInstance((_) => {});
-				scopeCache[model] = burden;
-				return burden;
+				scopedBurden = createInstance((_) => {});
+				scopeCache[model] = scopedBurden;
+				return scopedBurden;
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


generik0:
@jonorossi can you add netfore3 and 3.1 to the appveyor? Then the tests can run in all frameworks

generik0:
@AntonioDrusin can you review. it was the scopeCache that needed the lock. Hence we only sync for each scope. less performance hit, and more SOLID as scopeCache is an instance, not a static..

```c#
lock (scopeCache)
			{
				var burden = scopeCache[model];
				if (burden == null)
				{
					scopeCache[model] = burden = createInstance((_) => {});
				}
				return burden;	
			}
```

src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.2;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.2;netcoreapp3.0;netcoreapp3.1;net461;net471;net48</TargetFrameworks>
```
jonorossi, line 4: What is this for? Why do we need to compile the test project on all these platforms, especially with .NET Framework 4.x being an in place upgrade so a machine can only have a single version. If you want to run the unit tests on multiple .NET Core versions that'll be handled through the AppVeyor config in the root of the repo and the build script.

src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
```diff
@@ -69,15 +69,14 @@ public Burden GetCachedInstance(ComponentModel model, ScopedInstanceActivationCa
 				scopeCache[burden] = burden;
```
jonorossi, line 69: You are locking access to `scopeCache` below, but what about this code that is nearly a duplicate. Maybe refactor it?

jonorossi (changes requested):

src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
jonorossi: The AppVeyor build is also failing because it doesn't have .NET Core 3.0 installed, maybe need to pick a different AppVeyor image?

generik0: > What is this for? Why do we need to compile the test project on all these platforms, especially with .NET Framework 4.x being an in place upgrade so a machine can only have a single version. If you want to run the unit tests on multiple .NET Core versions that'll be handled through the AppVeyor config in the root of the repo and the build script.

Well before it was testing net461 and netcoreapp2.2. I can revert. This is how i normally Do multiple framework testing. It works very Well.
In this case, my thought was, "the issue was specific to netcoreapp3.1, then we should also run the tests for netcoreapp3.1"
If you have a better way to do it, I can revert to net461 and netcoreapp2.2 again?

src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
generik0: > You are locking access to `scopeCache` below, but what about this code that is nearly a duplicate. Maybe refactor it?

I will look. Yes there are issues with the original code that still needs to be cleaned up.

src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
jonorossi: It appears you are right that this is the way it is done so `dotnet test` and Visual Studio work, even though the test project get recompiled for each target. Go for it with the .NET Core targets, but I don't see any reason to add more .NET Framework targets since they are all the same runtime.

You'll likely have to update the AppVeyor config YAML file to get newer .NET Core runtimes on the build machine.

generik0: @jonorossi I don't see any references to the netcore versions available in the appveyor.yml I can change to the VS2019 image, but that is all. Only you have login to AppVeyor?

src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
generik0: @jonorossi please look at my change. If I need to refactor I will have to lock it all:

It appears that this is how the scopes are being managed by the facility. It isn't using Castle Scoping directly, I think this was a sort of wrapper.
Is this the correct way to go? and thoughts @jonorossi @AntonioDrusin 


```c#
public Burden GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
		{
			lock (scopeCache)
			{
				Burden burden = null;
				if(model.Configuration.Attributes.Get(TransientMarker) != bool.TrueString )
				{
					burden = scopeCache[model];
				}
				if (burden != null) return burden;
				scopeCache[model] = burden = createInstance((_) => {});
				return burden;
			}
		}
```

AntonioDrusin:
The lock code looks correct. I tested with @mdavis test project and with my own. The resolvers now behave as expected.

src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
```diff
@@ -63,20 +63,21 @@ public void Dispose()
...
 				}
-
+				else
+				{
+					burden = createInstance((_) => {});
+					scopeCache[burden] = burden;
+				}
+				if (burden != null) return burden;
+				burden = createInstance((_) => {});
+				scopeCache[model] = burden;
```
jonorossi, line 80: There is some duplication in this method, maybe go with this:

```csharp
		public Burden GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
		{
			lock (scopeCache)
			{
				Burden burden = null;

				if (model.Configuration.Attributes.Get(TransientMarker) != bool.TrueString)
				{
					burden = scopeCache[model];
				}

				if (burden == null)
				{
					burden = createInstance((_) => {});
					scopeCache[burden] = burden;
				}

				return burden;
			}
		}
```

src/Castle.Windsor.Extensions.DependencyInjection.Tests/Castle.Windsor.Extensions.DependencyInjection.Tests.csproj
jonorossi: VS2019 AppVeyor image is the right fix, AppVeyor has the SDKs installed on the images.

src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
jonorossi: Maybe this would make it clearer what is happening:

```csharp
		public Burden GetCachedInstance(ComponentModel model, ScopedInstanceActivationCallback createInstance)
		{
			lock (scopeCache)
			{
				Burden burden = null;

				// Add transient's burden to scope so it gets released
				if (model.Configuration.Attributes.Get(TransientMarker) == bool.TrueString)
				{
					burden = createInstance((_) => {});
					scopeCache[burden] = burden;
					return burden;
				}

				burden = scopeCache[model];
				if (burden == null)
				{
					burden = createInstance((_) => {});
					scopeCache[model] = burden;
				}
				return burden;
			}
		}
```

generik0:
@jonorossi improved nestling and improved naming of variables for readability

generik0:
@jonorossi did i get all the review comments ok? If so, can we get this into master and a new build for people?

jonorossi (approved):

jonorossi:
@generik0 yep, all good. Merged to master.

PR #551, opened 2020-10-05
Fix components being cleared when registering existing container in host

### LINKED ISSUE #550: Extensions.DependencyInjection: Existing services in container are cleared when registering in host
The use case is creating a container and registering components before registering it with the host. The naming sub-system is replaced, and all existing components are cleared.

The following example fails:

```c#
public interface IFileSystem
{
}

public class FileSystem : IFileSystem
{
}

public interface ICustomFileWriter
{
}

public class CustomFileWriter : ICustomFileWriter
{
	public CustomFileWriter(IFileSystem fileSystem)
	{
	}
}

public class Startup
{
	public void Configure(IApplicationBuilder app, ICustomFileWriter writer)
	{
            // ICustomFileWriter fails to resolve
	}
}

public class Program
{
	public static void Main(string[] args)
	{
		var container = new WindsorContainer();
		container.Register(Component.For<IFileSystem>().ImplementedBy<FileSystem>());

		Host
			.CreateDefaultBuilder()
                        // Register our own container
			.UseWindsorContainerServiceProvider(container)
			.ConfigureServices(x => x.AddSingleton<ICustomFileWriter, CustomFileWriter>())
			.ConfigureWebHostDefaults(x => x.UseStartup<Startup>())
			.Build()
			.Run();
	}
}
```

robertcoltheart:
Fixes #550 

My quick and dirty fix is just to copy the existing handlers from the old naming subsystem to the new one. Not sure if its the right fix, but hopefully this illustrates the issue.

### DIFF
--- a/src/Castle.Windsor/MicroKernel/DefaultKernel.cs
+++ b/src/Castle.Windsor/MicroKernel/DefaultKernel.cs
@@ -342,7 +342,17 @@ namespace Castle.MicroKernel
 			}
 			else if (name == SubSystemConstants.NamingKey)
 			{
-				NamingSubSystem = (INamingSubSystem)subsystem;
+				var namingSubSystem = subsystem as INamingSubSystem;
+
+				if (NamingSubSystem != null)
+				{
+					foreach (var handler in NamingSubSystem.GetAllHandlers())
+					{
+						namingSubSystem?.Register(handler);
+					}
+				}
+
+				NamingSubSystem = namingSubSystem;
 			}
 		}
 
--- a/src/Castle.Windsor.Extensions.DependencyInjection.Tests/Components/EmptyService.cs
+++ b/src/Castle.Windsor.Extensions.DependencyInjection.Tests/Components/EmptyService.cs
@@ -1,0 +1,20 @@
+﻿// Copyright 2004-2020 Castle Project - http://www.castleproject.org/
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.Windsor.Extensions.DependencyInjection.Tests.Components
+{
+	public class EmptyService : IEmptyService
+	{
+	}
+}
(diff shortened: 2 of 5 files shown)


src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -342,7 +342,17 @@ public virtual void AddSubSystem(String name, ISubSystem subsystem)
...
 			else if (name == SubSystemConstants.NamingKey)
 			{
-				NamingSubSystem = (INamingSubSystem)subsystem;
+				var namingSubSystem = subsystem as INamingSubSystem;
+
+				if (NamingSubSystem != null)
+				{
+					foreach (var handler in NamingSubSystem.GetAllHandlers())
+					{
+						namingSubSystem?.Register(handler);
```
generik0, line 351: @robertcoltheart we cannot use c notation in Castle code standards, you need to if(namingSubSystem!=null)

generik0 (changes requested):
Tests do not pass when using this option.

```c#
public class WindsorServiceProviderFactoryTests : DependencyInjectionSpecificationTests
	{
		
		protected override IServiceProvider CreateServiceProvider(IServiceCollection serviceCollection)
		{
			using (var windsorContainer = new WindsorContainer())
			{
				windsorContainer.Register(Component.For<IEmptyService>().ImplementedBy<EmptyService>());

				var factory = new WindsorServiceProviderFactory(windsorContainer);
				var container = factory.CreateBuilder(serviceCollection);
				return factory.CreateServiceProvider(container);
			}
		}
	}
```

image

robertcoltheart:
Yeah this was really just a starting point to illustrate the problem. See my comments on #550, and if there is a better way to deal wit this, I'm happy to modify this PR.

generik0:
I like your PR because it means we do not have to do this:
image

src/Castle.Windsor/MicroKernel/DefaultKernel.cs
```diff
@@ -342,7 +342,17 @@ public virtual void AddSubSystem(String name, ISubSystem subsystem)
...
+					}
```
jonorossi, line 352: The problem with this is that a naming subsystem is more than just handlers, this doesn't account for handler selectors and filters. Even though the MicroKernel (part of Windsor now) exposes `INamingSubSystem` and `AddSubSystem` is definitely hasn't been an extension point that has been used much or at all so it is possible we'll run into other areas where things don't work quite right.

Could we achieve the ordering of handlers somewhere else rather than a custom naming subsystem?

jonorossi (changes requested):

robertcoltheart:
I don't know the inner workings of Windsor well enough to make those kinds of changes, so I'll park this for now. I have a viable workaround, so I'm happy to let this go.

PR #553, opened 2020-10-05
Update ServiceDescriptorExtensions.cs

twenzel:
- Improve performance due to less reflection
- Prevent runtime errors when renaming RegistrationAdapter.FromServiceDescriptor method

### DIFF
--- a/src/Castle.Windsor.Extensions.DependencyInjection/Extensions/ServiceDescriptorExtensions.cs
+++ b/src/Castle.Windsor.Extensions.DependencyInjection/Extensions/ServiceDescriptorExtensions.cs
@@ -14,14 +14,10 @@
 
 namespace Castle.Windsor.Extensions.DependencyInjection.Extensions
 {
-	using System.Reflection;
-
 	using Castle.MicroKernel.Registration;
 
 	public static class ServiceDescriptorExtensions
 	{
-		private static MethodInfo fromServiceDescriptorMethod = typeof(RegistrationAdapter).GetMethod(nameof(RegistrationAdapter.FromServiceDescriptor), BindingFlags.Static | BindingFlags.Public);
-
 		public static IRegistration CreateWindsorRegistration(this Microsoft.Extensions.DependencyInjection.ServiceDescriptor service)
 		{
 			if (service.ServiceType.ContainsGenericParameters)
@@ -30,7 +26,7 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Extensions
 			}
 			else
 			{
-				var method = fromServiceDescriptorMethod.MakeGenericMethod(service.ServiceType);
+				var method = RegistrationAdapterMethods.FromServiceDescriptor.MakeGenericMethod(service.ServiceType);
 				return method.Invoke(null, new object[] { service }) as IRegistration;
 			}
 		}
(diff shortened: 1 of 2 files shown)


ltines:
Nice @twenzel. Thanks

robertcoltheart:
Why not just use a generic type registration? I've raised #555 to consider instead of this.

src/Castle.Windsor.Extensions.DependencyInjection/Extensions/ServiceDescriptorExtensions.cs
```diff
@@ -20,6 +20,8 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Extensions
 
 	public static class ServiceDescriptorExtensions
 	{
+		private static MethodInfo fromServiceDescriptorMethod = typeof(RegistrationAdapter).GetMethod(nameof(RegistrationAdapter.FromServiceDescriptor), BindingFlags.Static | BindingFlags.Public);
```
jonorossi, line 23: If we do go ahead with this rather than #555, then this should be marked readonly and probably follow the pattern we use throughout Castle.

e.g.:

```csharp
/// <summary>
/// Holds <see cref="MethodInfo" /> objects representing methods of the <see cref="RegistrationAdapter" /> class.
/// </summary>
internal static class RegistrationAdapterMethods
{
	public static readonly MethodInfo FromServiceDescriptor =
		typeof(RegistrationAdapter).GetMethod(nameof(RegistrationAdapter.FromServiceDescriptor), BindingFlags.Static | BindingFlags.Public);
}
```

jonorossi (changes requested):

src/Castle.Windsor.Extensions.DependencyInjection/Extensions/ServiceDescriptorExtensions.cs
twenzel: Done.

jonorossi (approved):

generik0:
Hi @jonorossi i agree with this change

jonorossi:
@generik0 you've commented both here and on #555 that you agree with the change, isn't it one or the other?

generik0:
@jonorossi i did not see changes on #555 and #553 to be mutually exclusive.
This is to improve performance on a method call:

```c#
internal static class RegistrationAdapterMethods
	{
		public static readonly MethodInfo FromServiceDescriptor =
			typeof(RegistrationAdapter).GetMethod(nameof(RegistrationAdapter.FromServiceDescriptor), BindingFlags.Static | BindingFlags.Public);
	}
```

The other to remove some reflections.
```c#
public static IRegistration FromServiceDescriptor(Microsoft.Extensions.DependencyInjection.ServiceDescriptor service)
		{			{
			var registration = Component.For<TService>()				var registration = Component.For(service.ServiceType)
				.NamedAutomatically(UniqueComponentName(service));					.NamedAutomatically(UniqueComponentName(service));
```


Do the changes not complement each other? @twenzel @robertcoltheart

robertcoltheart:
I mean, reflection is always going to be slow. I haven't benchmarked either change to see which is faster though.

generik0:
@twenzel @robertcoltheart I think we should remove reflections. But that is just an opinion. Maybe you guys can decide?

jonorossi:
Closing as we'll go with #555 which puts the implementation in line with `FromOpenGenericServiceDescriptor`. Thanks for the PR @twenzel.

PR #555, opened 2020-10-06
Use generic registration instead of reflection

robertcoltheart:
Components will be of type `Component<object>` instead. This would supersede the change in https://github.com/castleproject/Windsor/pull/553.


generik0:
@jonorossi I agree with this change :-)

Issue #552, opened 2020-10-05
Decorators does not work anymore (5.1.0-beta001)


twenzel:
After upgrading from v4.x to 5.1-beta I found that the decorators does not work anymore.

**Exception:**
```
HandlerException: Can't create component 'WebApplication3UsingWindsor.DecoratedUserService' as it has dependencies to be satisfied.

'WebApplication3UsingWindsor.DecoratedUserService' is waiting for the following dependencies:
- Service 'WebApplication3UsingWindsor.IUserService' which points back to the component itself.
A dependency cannot be satisfied by the component itself, did you forget to register other components for this service?
```

Using: Castle.Windsor.Extensions.DependencyInjection (5.1.0-beta001)

In previous version this was working:
```Csharp
container.Register(Component.For<IUserService>().ImplementedBy<DecoratedUserService>());
...
container.Register(Component.For<IUserService>().ImplementedBy<UserService>());
```

Now only this is working:
```Csharp
container.Register(
			Component.For<IUserService>().ImplementedBy<DecoratedUserService>(),
			Component.For<IUserService>().ImplementedBy<UserService>()
			);
```

I don't want to re-write the whole register logic in my application and try to find all occurences.

Is this new behavior by design?

I've attached a simple repro project.
[WebApplication3UsingWindsor.zip](https://github.com/castleproject/Windsor/files/5329218/WebApplication3UsingWindsor.zip)

twenzel:
I think the error comes from `Castle.Windsor.Extensions.DependencyInjection`.
Using `Castle.Windsor.MsDependencyInjection` with `Castle.Windsor` (5.1.0-beta001) works like a charm.

generik0:
HI @ltines this seams to what is causing the issue;:

![image](https://user-images.githubusercontent.com/2411563/95330273-93ac4280-08a8-11eb-84e4-76682ccc7934.png)
I though @robertcoltheart PR #551 but it doesn't.

@ltines can you tell us the reason for the "AddSubSystem" in the facility? and also, can you tell us why this might be giving an issue?
Thanks :-)

generik0:
@ltines this will fix the issue;
Any thought to improving?

![image](https://user-images.githubusercontent.com/2411563/95332682-0ec32800-08ac-11eb-9410-8f1ab51fa0cb.png)

generik0:
@jonorossi @ltines @twenzel @AntonioDrusin @robertcoltheart Got it.. Anyone have time to comment / review?
#556

[...]

PR #556, opened 2020-10-07
Add InvalidateCache to DependencyInjectionNamingSubsystem

generik0:
Add InvalidateCache to DependencyInjectionNamingSubsystem when services are registering

#552

### DIFF
--- a/src/Castle.Windsor.Extensions.DependencyInjection.Tests/SubSystems/DependencyInjectionNamingSubsystemTests.cs
+++ b/src/Castle.Windsor.Extensions.DependencyInjection.Tests/SubSystems/DependencyInjectionNamingSubsystemTests.cs
@@ -23,42 +23,58 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Tests.SubSystems
 	using Xunit;
 
 	public class DependencyInjectionNamingSubsystemTests
-    {
-        [Fact]
-        public void Can_Decoration_Resolve_When_Named()
-        {
+	{
+		[Fact]
+		public void Can_Decoration_Resolve_When_Named()
+		{
 			var container = new WindsorContainer();
-			
+
 			var target = new DependencyInjectionNamingSubsystem();
 			container.Kernel.AddSubSystem(SubSystemConstants.NamingKey, target);
 
-	        container.Register(Component.For<IUserService>().ImplementedBy<DecoratedUserService>().Named(nameof(DecoratedUserService)), 
-		        Component.For<IUserService>().ImplementedBy<UserService>().Named(nameof(UserService))
-	        );
+			container.Register(Component.For<IUserService>().ImplementedBy<DecoratedUserService>().Named(nameof(DecoratedUserService)),
+				Component.For<IUserService>().ImplementedBy<UserService>().Named(nameof(UserService))
+			);
 
-	        var actualDecoratedUserService = container.Resolve<IUserService>(nameof(DecoratedUserService));
+			var actualDecoratedUserService = container.Resolve<IUserService>(nameof(DecoratedUserService));
 			Assert.NotNull(actualDecoratedUserService);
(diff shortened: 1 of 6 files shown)


src/Castle.Windsor.Extensions.DependencyInjection/SubSystems/DependencyInjectionNamingSubsystem.cs
```diff
@@ -123,5 +123,11 @@ public override IHandler GetHandler(Type service)
 			return null;
 		}
 
+		protected override void InvalidateCache()
+		{
+			handlerListsRegistrationOrderByTypeCache.Clear();
+			handlerListsPriorityOrderByTypeCache.Clear();
+			base.InvalidateCache();
+		}
```
jonorossi, line 131: Can you add a unit test to show what this is fixing and so we don't get that defect back.

jonorossi (changes requested):

src/Castle.Windsor.Extensions.DependencyInjection/SubSystems/DependencyInjectionNamingSubsystem.cs
generik0: Yes, but not until next week.
Good point, Thabks

generik0:
Hi @jonorossi. Tests added. Also cleaned up unused collection. I have changed the handlerListsByTypeCache to be protected, so the overriding GetHandler uses the same collection as the base, And hence the InvalidateCache in the base does not need to be overridden.
Is this as to be expected? @ltines @AntonioDrusin @robertcoltheart @twenzel and thoughts/comments?

src/Castle.Windsor.Extensions.DependencyInjection.Tests/SubSystems/DependencyInjectionNamingSubsystemTests.cs
```diff
@@ -0,0 +1,64 @@
...
+	        container.Register(Component.For<IUserService>().ImplementedBy<DecoratedUserService>(),
+		        Component.For<IUserService>().ImplementedBy<UserService>()
+	        );
+
+	        var actual = container.Resolve<IUserService>();
+	        Assert.NotNull(actual);
+	        Assert.IsType<DecoratedUserService>(actual);
+	        Assert.NotNull(((DecoratedUserService)actual).UserService);
+        }
+
```
twenzel, line 62: Please add following test also:

```csharp
[Fact]
        public void Can_Decorator_Resolve_Decorated_Class_When_Registered_Consecutively()
        {
	        var container = new WindsorContainer();

	        var target = new DependencyInjectionNamingSubsystem();
	        container.Kernel.AddSubSystem(SubSystemConstants.NamingKey, target);

	        container.Register(Component.For<IUserService>().ImplementedBy<DecoratedUserService>() );

 	        container.Register( Component.For<IUserService>().ImplementedBy<UserService>());

	        var actual = container.Resolve<IUserService>();
	        Assert.NotNull(actual);
	        Assert.IsType<DecoratedUserService>(actual);
	        Assert.NotNull(((DecoratedUserService)actual).UserService);
        }
```

generik0:
Thanks @twenzel ypu are welcome to push to my fork if you want.
Otherwise i can fix in a couple of days.

twenzel:
@generik0 Sure I can. Can you provide me some permissions.
Let's talk per twitter DM (@twenzel) or Slack (twenzel)

generik0:
Hi @twenzel I don't think the fork is locked is it? Other have been able to push to it. Is you try?
I don't have Twitter, and slack I do not believe you can communicate directly?
You can email and we will sort it all out. rsr@gait.dk

generik0:
@jonorossi if you have time, looks like we are ready to complete :-)

jonorossi (approved):

Issue #557, opened 2020-10-07
Code error in AddSubSystemToContainer, wrong container


generik0:
Should be using container, not rootContainer

```c#
protected virtual void AddSubSystemToContainer(IWindsorContainer container)
		{
			rootContainer.Kernel.AddSubSystem(
				SubSystemConstants.NamingKey,
				new SubSystems.DependencyInjectionNamingSubsystem()
			);
		}
```

PR #558, opened 2020-10-07
Should use the container from the method call, not the root container

generik0:
Should use the container from the method call, not the root container

#557


robertcoltheart (approved):

jonorossi (commented):
Generally `WindsorServiceProviderFactoryBase` feels very fragile, it has virtual functions that expose each implementation detail so you can override whichever bits you'd like, looks like you'd have to read the code of this base class to know what to override.

PR #571, opened 2020-11-25
Feature/565 remove castle core lock

generik0:
Removed usage of Lock from Castle.Core in benefit for ReaderWriterLockSlim

### DIFF
--- a/src/Castle.Facilities.WcfIntegration/Client/Discovery/DiscoveryProxyCache.cs
+++ b/src/Castle.Facilities.WcfIntegration/Client/Discovery/DiscoveryProxyCache.cs
@@ -16,13 +16,15 @@ namespace Castle.Facilities.WcfIntegration
 {
 	using System;
 	using System.ServiceModel.Discovery;
+	using System.Threading;
+
 	using Castle.Core.Internal;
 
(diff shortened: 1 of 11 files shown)


generik0:
HI @jonorossi i needed to change the Dispose in the CallContextLifetimeScope to use the disposal pattern, instead of locks. The old Lock apparently did not lock for release. (or something)...
Do you agree with the change,, or should I go a different direction?

jonorossi (changes requested):
Wow okay, that really increased the lines of code and complexity around locking.

The reason we deprecated these locking classes in Castle Core is that Castle Core needs to stop being the old dumping ground for utilities since the Castle Project really is just DynamicProxy and Windsor now. Versioning all these public classes there where they don't even get used makes no sense.

Maybe we should add a `Lock` class to Windsor instead to keep the ability to use a `using` statement which is neat. Do you want to look at adding a `Lock` class (https://github.com/castleproject/Core/tree/v4.4.1/src/Castle.Core/Core/Internal) without all the interfaces and additional stuff needed to support Silverlight?

generik0:
@jonorossi i woke up thinking the exactly same thing ;-)

Issue #564, opened 2020-11-18
Castle.Core target frameworks vs Castle.Windsor


generik0:
Hi @jonorossi 

The Castle.Core target references are:
`<TargetFrameworks>net45;netstandard2.0;netstandard2.1</TargetFrameworks>`

Should we align Castle.Windsor against Castle.Core?

Even better:

`<TargetFrameworks>net45;netstandard2.0;net5.0</TargetFrameworks>`

?

generik0:
After a bit of local testing. Net45 and netstandard1.6 does not build.


So the question is, should we add 2.1 to the build for

jonorossi:
I think this one will need to wait a bit. Castle.Core target frameworks in master are newer than the last release, we've not yet got to release Castle.Core 5.0.

- 4.4.1: `net35;net40;net45;netstandard1.3;netstandard1.5`
- master: `net45;netstandard2.0;netstandard2.1`

Now that `net5.0` is out, we'll add that and probably drop ns2.1.

generik0:
@jonorossi please review PR whne you have time. Castle.Windsor looks ready to update target frameworks

lg2de:
> Now that `net5.0` is out, we'll add that and probably drop ns2.1.

netstandard2.1 is used if referenced from dotnetcoreapp3.1. If you remove ns2.1 and add net5.0 you will not support dotnetcore 3.1. But 3.1 is LTS, net5.0 is NOT LTS.

jonorossi:
> If you remove ns2.1 and add net5.0 you will not support dotnetcore 3.1. But 3.1 is LTS, net5.0 is NOT LTS.

Windsor today targets `net45;netstandard1.6`, so "dropping" ns2.1 in my comment was from the proposed list which @generik0 posted (since it isn't implemented), which would leave the new proposed list `net45;netstandard2.0;net5.0`.

Anyway, these comments are old compared to the discussion on the pull request.

PR #572, opened 2020-11-25
Change target framework from .NET Standard 1.6 to 2.0, and move .NET Core tests to .NET 5

generik0:
#564


jonorossi:
The build fails.

buildscripts/BuildScripts.csproj
```diff
@@ -1,18 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>net461;netcoreapp1.1</TargetFrameworks>
+		<TargetFramework>net5.0</TargetFramework>
```
generik0, line 4: @jonorossi is this a good change, I don't see the need for the old stuff. Gavin added to reflect "look and feel" of castle.core.

generik0:
@jonorossi Ready. Change in test framework and those nasty preprocessor  labels we have in windsor :-/

buildscripts/BuildScripts.csproj
jonorossi: It would have likely also been to support Mono at the time.

I assume this sets the default for packages, seems like a decent default, however would we be better off with no default?

src/Castle.Windsor/Castle.Windsor.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>net45;netstandard1.6</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard2.0;net5.0</TargetFrameworks>
```
jonorossi, line 4: In summary, we are dropping .NET Standard 1.6 and adding .NET Standard 2.0 and .NET 5.0, correct?

.NET Standard 2.0 for .NET Framework 4.6.1+, .NET Core 2.x and .NET Core 3.x. .NET Core 1.x is unsupported now anyway.

buildscripts/BuildScripts.csproj
generik0: I think Gavin just copied from caste.core. The project is only a wrapper for some scripts, there is nothing to build in the project, if I unload, rebuild is still green.

Should I remove in favour of a folder?

src/Castle.Windsor/Castle.Windsor.csproj
generik0: I read this on a MS page. I.e. that this was the recommendation (of cause right now, I cannot find again)
net45 to net46 = net45
net461 to net48 and netappcore2.0 to netapp3.1 = netstandard2.0
net5.0 = net 5.0

this will mean ofcause that netapp1.x will not be able to update.

buildscripts/BuildScripts.csproj
jonorossi: It might have been to keep the files in `BuildScripts.csproj` in the solution.

src/Castle.Windsor/Castle.Windsor.csproj
jonorossi: https://docs.microsoft.com/en-us/dotnet/standard/net-standard

What value does adding a `net5.0` target add? .NET 5 supports .NET Standard 2.0 so it should be unnecessary right? And is just increasing build and test time, and package size?

jonorossi:
There is a few of these warnings in the build log, should we be doing something about them?

> C:\Program Files\dotnet\sdk\5.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp2.0' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.

buildscripts/BuildScripts.csproj
generik0: We can just use a solution folder. with a nice name "BuidlScripts".. Shoud i change?

src/Castle.Windsor/Castle.Windsor.csproj
generik0: I read the MS recommends net5.0 targetting for packaging. 
Don't you think, making a net5.0 specific build, will build the packages with the specific improvements from the net5.0 sdk and increase e.g. increase runtime performance for net5 applications (or am I just dreaming)?

I can delete if you insist :-)

generik0:
> There is a few of these warnings in the build log, should we be doing something about them?
> 
> > C:\Program Files\dotnet\sdk\5.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp2.0' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.

The netcoreapp2.0 issue is from the Castle.Facilities.AspNetCore tests. The Castle.Facilities.AspNetCore as you know is only for netcoreapp2.0, and the netcoreapp2.0 tests

The netcoreapp2.2 is from the new dependency injection extension tests. I have removed, but I actually thought it was nice to show that netcoreapp2.2 is supported with the dependency injection extension.

My question is. Can we not remove the Castle.Facilities.AspNetCore and Castle.Facilities.AspNetCore.Tests. Are we making new builds of this? If people want to use this facility should they not use Windsor 5.0? I guess this raises the old question if the old asp.net, wcf and now alos aspnetcore facility shouldn't be moved to an obsolete repo. But I think you should decide here :-)


As with the build issue. I can add the warnings to the build ignore.? or we can leave it. What is best?

buildscripts/BuildScripts.csproj
jonorossi: Sure, go for it as long as that's "BuildScripts" 😉

src/Castle.Windsor/Castle.Windsor.csproj
jonorossi: I don't insist on anything, I'd just like to understand the guidance from the .NET team as I probably know less about .NET 5 than you do. I do recall reading somewhere that .NET Standard is now seen as "mature" (i.e. pretty much end of life) and that the `net5.0` target is the future, but wasn't sure whether that was the official position or .NET Standard libraries are still good. I can't see there being any difference in performance over which TFM you choose as it is just an API surface and still the same runtime.

[The future of .NET Standard](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/) blog post on the .NET Blog talks about the difference. My understanding is that offering a `net5.0` TFM will make no difference since we will also have a `netstandard2.0` one (.NET 5 supports all .NET Standard versions), until we want to use .NET 5 features which this will enable.

I'm happy for this to include the `net5.0` if you think you'll want to use features in .NET 5, but those will have to be optional to keep support for older TFMs.

If anyone else reading along has more info please jump in.

generik0:
Hi @stakx and  @Lukas Grützmacher we are looking into moving castle.core's target frameworks to Castle.Windsor.
Can you tell us why do you build Castle.Core for both ;netstandard2.0 and netstandard2.1. and not just ;netstandard2.0? Is there a technical reason? Thnaks for your time

generik0:
@jonorossi once we get all the details cleared up, i think i will create a new PR with less commits. A more clean/correct PR / Branch

lg2de:
@generik0 The library Castle.Core may not need to build for netstandard2.1 because it does not use new features.
BUT, other projects refer to this base library. Depending on its own target it needs to refer "compatibility packages" for new function not available in netstandard2.0.

I run into problems with the limited list of target framework because we have an internal approval process for packages from nuget.org. The more packages are referenced indirectly the more packages we need to approve.
Please do not ask me at the moment for more details. I've analyzed this months ago. I guess it was related to the package "Moq".

I recommend to target netstandard2.0 AND netstandard2.1 for more flexibility to all users.

stakx:
@generik0:

> Can you tell us why do you build Castle.Core for both ;netstandard2.0 and netstandard2.1. and not just ;netstandard2.0?

This was decided in https://github.com/castleproject/Core/issues/407#issuecomment-620601558. I'll try to summarize from what I remember: I argued that we should target `netstandard2.1`, since `netstandard2.0` doesn't officially support System.Reflection.Emit, which DynamicProxy heavily depends on. @jonorossi agreed that this is the .NET Standard version we want upstream code to target in the longer term, but felt that we should still have `netstandard2.0` (I suppose in order to have .NET Core 2.x covered) because in practice, System.Reflection.Emit will still work just fine; so the "missing official support" argument by itself is not good enough to exclude `netstandard2.0`.

(Btw. while we're on the topic of target frameworks, re: `net5.0` &ndash; there has been an [announcement that .NET Standard 2.1 will be the last version of .NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#when-to-target-net50-vs-netstandard). Fortunately, .NET 5 implements `netstandard2.1` so as long as you don't require any new platform or language features, you shouldn't have to add `net5.0` as a target framework.)

lg2de:
> so as long as you don't require any new platform or language features, you shouldn't have to add `net5.0` as a target framework.

With the same arguments as above I vote for adding net5.0 in addition to other targets. Building libraries on top of Castle (Core), the user has more flexibility to reference what is required for each target.

generik0:
@jonorossi I think you need to decide. I have no preference for any of it :-)

However i do feel we should mirror Castle.Core and make tests run to each included framework. Hence for now make Castle.Windsor use .net45, .netstandard2.0 and .netstandard2.1. and the tests .net472, and .net5.0.

In the future --> if Castle.Core add a framework or Castle.Windsor needs something specific from a framework then add this also.

jonorossi:
> Can we not remove the Castle.Facilities.AspNetCore

I have no problem removing the obsolete facilities now that we've got a proper Extensions DependencyInjection one. Those that are stuck on old versions of ASP.NET Core won't need a new Windsor anyway. Log an issue to everyone following alone will see the notification.

> once we get all the details cleared up, i think i will create a new PR with less commits. A more clean/correct PR / Branch

Just use rebase to clean up your commits on the branch, GitHub will detect the changes and automatically update this pull request. Or I'll just squash everything together before merging.

> Can you tell us why do you build Castle.Core for both ;netstandard2.0 and netstandard2.1. and not just ;netstandard2.0?

As @stakx explained, there is annoying problems with reflection emit not properly being exposed with specific .NET Standard versions. We can do something different here with Windsor.

> With the same arguments as above I vote for adding net5.0 in addition to other targets

@lg2de if Castle Core doesn't target `net5.0` yet I assume it won't help you reduce the number of dependencies?

> However i do feel we should mirror Castle.Core and make tests run to each included framework. Hence for now make Castle.Windsor use .net45, .netstandard2.0 and .netstandard2.1. and the tests .net472, and .net5.0.
> 
> In the future --> if Castle.Core add a framework or Castle.Windsor needs something specific from a framework then add this also.

We could change Windsor from `net45;netstandard1.6` to just `netstandard2.0`, that'd give us support for .NET Core 2.0+ and .NET Framework 4.6.1+, I see no need to support less than .NET Framework 4.6.1 at this point, there are always older versions of Windsor people can use. Then once Castle Core adds `net5.0` we can look at adding it here too if it helps simplify dependencies.

lg2de:
> @lg2de if Castle Core doesn't target `net5.0` yet I assume it won't help you reduce the number of dependencies?

The less dependencies the better!
A library which is used quite often in other libraries should be provided with minimal and exact matching target framework.

I do not understand your question in relation to net5.0.

jonorossi:
@lg2de you said:
> I run into problems with the limited list of target framework because we have an internal approval process for packages from nuget.org. The more packages are referenced indirectly the more packages we need to approve.
Please do not ask me at the moment for more details. I've analyzed this months ago. I guess it was related to the package "Moq".

and:
> With the same arguments as above I vote for adding net5.0 in addition to other targets. Building libraries on top of Castle (Core), the user has more flexibility to reference what is required for each target.

You are requesting something but providing little information, I attempted to read between the lines but obviously was wrong. I thought you were saying that under `net5.0` TFM it has less corefx libraries or something, we definitely know that .NET Standard 1.x TFMs have heaps of implicit dependencies, adding .NET Standard 2.x will fix that problem. Can you please explain why it would help you if we add the `net5.0` TFM.

generik0:
Hi, @jonorossi looks like the PR is good then. 
Change frameworks to **core net45;netstandard2.0** and **tests net45;net5.0**
Should we start her, And Continue the discussion for other changes in #564

lg2de:
@jonorossi, I did not request anything here. I was asked for details on discussion months ago.
I tried to remember. Unfortunately I do NOT get back all the details.
I think it was a complex constellation with multiple libraries referencing over several steps causing unexpected high number of packages finally referenced (directly and indirectly) in MY project.

At the time of my analysis net5.0 was not available, or at least not in my focus.

> Can you please explain why it would help you if we add the `net5.0` TFM.

At the moment I can not.
I was focusing on Castle.Core only, not Windsor. So, please proceed with the information you have.
As soon I hit the problem again, I'll reanalyze again with the current situation.

generik0:
Yes, I looked in the Castle.Core history, and tried to discover why there was netstandard2.1 and asked for some historical information:-)

lg2de:
> Yes, I looked in the Castle.Core history, and tried to discover why there was netstandard2.1 and asked for some historical information:-)

... and stakx gave the correct answer I did not remember :(

jonorossi:
@lg2de I understood that part and knew exactly why Castle Core has .NET Standard 2.0 and 2.1. I was only asking about this comment of yours because that is what we are discussing, if we should add `net5.0` or not:

> With the same arguments as above I vote for adding net5.0 in addition to other targets.

lg2de:
At the moment I do not have strong arguments to add net5.0.

generik0:
I think we are good with "Change frameworks to net45;netstandard2.0 and tests net45;net5.0 where applicable"

src/Castle.Windsor/Castle.Windsor.csproj
```diff
@@ -33,10 +33,18 @@
 		<Reference Include="System.Web" />
 	</ItemGroup>
 
-	<ItemGroup Condition="'$(TargetFramework)'=='netstandard1.6'">
-		<PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
-		<PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
-		<PackageReference Include="System.Threading.Thread" Version="4.3.0" />
-		<PackageReference Include="Microsoft.Extensions.DependencyModel" Version="1.1.2" />
-	</ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0' OR '$(TargetFramework)'=='net5.0'">
```
jonorossi, line 36: No need for the `net5.0` conditional here.

src/Castle.Windsor/Castle.Windsor.csproj
```diff
@@ -33,10 +33,18 @@
...
+    <PackageReference Include="System.Threading.Thread" Version="4.3.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'=='net5.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="5.0.0" />
+  </ItemGroup>
```
jonorossi, line 48: No need for the `net5.0` conditional here.

jonorossi (changes requested):
Just a couple of minor comments.

src/Castle.Windsor/Castle.Windsor.csproj
```diff
@@ -33,10 +33,18 @@
...
-	</ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0' OR '$(TargetFramework)'=='net5.0'">
+    <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
+    <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+    <PackageReference Include="System.Threading.Thread" Version="4.3.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
+  </ItemGroup>
```
jonorossi, line 44: Might as well keep them all together like they were in .NET Standard 1.6.

generik0: You mean the version 1.1.2?

jonorossi: > Might as well keep them all together like they were in .NET Standard 1.6.
i.e. in the same `ItemGroup` like you've now done.

> You mean the version 1.1.2?
2.1.0 was probably the right version to use now since moving to .NET Standard 2.0.

jonorossi:
Thanks, merged.

Issue #563, opened 2020-11-18
Castle.Windsor.Extensions.DependencyInjection parallelism issues


nativenolde:
@ltines @generik0 thank you for contributing the Castle.Windsor.Extensions.DependencyInjection package. 
I guess a lot of people are very happy about that!

Since migrating my application to use the ms-di implementation (v5.1), i've noticed some potentially serious issues. They occur when running multiple ASP.NET Core Integration Tests in parallel on independent application instances. It seems some dependencies will be shared respectively disposed across HTTP-Request boundaries. At the moment, i can't say whether this is just a problem when testing or it's affecting production runtime as well.

My reproducible sample, created using Microsoft guidelines
https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests

is attached here
[InteractionTesting.zip](https://github.com/castleproject/Windsor/files/5559872/InteractionTesting.zip)

```
    public class Program
    {
        public static void Main(string[] args)
        {
            CreateHostBuilder(args)
                .Build()
                .Run();
        }

        public static IHostBuilder CreateHostBuilder(string[] args)
        {
            return Host.CreateDefaultBuilder(args)
                .UseWindsorContainerServiceProvider()
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.UseStartup<Startup>();
                });
        }
    }
```

```
    public class Startup
    {
        // This method gets called by the runtime. Use this method to add services to the container.
        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
        public void ConfigureServices(IServiceCollection services)
        {
            services
                .AddControllersWithViews()
                .AddControllersAsServices();
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app)
        {
            app.UseDeveloperExceptionPage();

            app.UseRouting();

            app.UseEndpoints(endpoints => { endpoints.MapDefaultControllerRoute(); });
        }
    }
```

```
    [Route("[controller]")]
    public class DemoController : ControllerBase
    {
        public IActionResult Demo()
        {
            return Ok();
        }
    }
```
As long as the tests are executed serial they work properly.. but if you run them in parallel, they fail for different reasons.
![image](https://user-images.githubusercontent.com/45496636/99528909-40e1a280-299f-11eb-8779-2b6017d8e4f9.png)
As an example:
```
2020-11-18 12:56:56 [Debug] Hosting starting
[...]

PR #573, opened 2020-11-25
SPIKE: Remove static scope handling

generik0:
This is a spike to allow parallel AppDomain usage of DI Extensions.
Needs to be cleaned up. 

Added tests for parrallel handling

**Do NOT MERGE TO MASTER**

#563


generik0:
- Statics removed in favor of singleton instance of ExtensionContainerScope.
- Instance of the ExtensionContainerScope is made shared between thread via AsyncLocal instance, Hence allowing root scope to be related to the initial assembely + thread it was created in
- Cleaned up, removing unneeded logic to begin a scope, as it is always related to the root or current scope

PR #594, opened 2021-03-23
Fix instantiate typo on component registration exception.

ChristianoKiss:
When raising ComponentRegistrationException, the message had a typo on the word "instantiate" (currently it is "instansiate").


jonorossi:
Thanks

PR #606, opened 2022-01-22
Update AbstractServiceHostBuilder.Generic.cs

gtame:
var service = model.Services.Single vs  var service = model.Services.FirstOrDefault();
The Single method throws exception when the collection has more one item. I have wrote my own ServiceHostBuilder to allow Service with multiple interfaces.
But this instruction don't allow me  to do.

   protected override void ValidateComponentModel(ComponentModel model)
        {
            //base.ValidateComponentModel(model);
        }
        protected override void ValidateServiceModel(ComponentModel model, ESPServiceModel serviceModel)
        {
          //  base.ValidateServiceModel(model, serviceModel);
        }

### DIFF
--- a/src/Castle.Facilities.WcfIntegration/Service/AbstractServiceHostBuilder.Generic.cs
+++ b/src/Castle.Facilities.WcfIntegration/Service/AbstractServiceHostBuilder.Generic.cs
@@ -79,7 +79,7 @@ namespace Castle.Facilities.WcfIntegration
 			ValidateComponentModel(model);
 			ValidateServiceModel(model, serviceModel);
 
-			var service = model.Services.Single();
+			var service = model.Services.FirstOrDefault();
 			foreach (var endpoint in serviceModel.Endpoints)
 			{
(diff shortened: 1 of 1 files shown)


src/Castle.Facilities.WcfIntegration/Service/AbstractServiceHostBuilder.Generic.cs
```diff
@@ -79,7 +79,7 @@ private void ValidateServiceModelInternal(ComponentModel model, TServiceModel se
 			ValidateComponentModel(model);
 			ValidateServiceModel(model, serviceModel);
 
-			var service = model.Services.Single();
+			var service = model.Services.FirstOrDefault();
```
jonorossi, line 82: Should this use `First` so maintain the requirement for non-null, or is null fine?

jonorossi (changes requested):

jonorossi:
Closing as stale. We also need a unit test for this change.

PR #613, opened 2022-05-14
Upgrade Castle.Core to 5.0.0 (resolves #612)

### LINKED ISSUE #612: Updating Windsor to support Castle.Core@5.0.0 and modern TFMs
With the release of Castle.Core@5.0.0, Windsor needs updating to support it as there are breaking changes (some detail in #610). 5.0.0 also has an updated list of TFMs:
```
<TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
```

The current targets for Windsor are:
```
Castle.Facilities.AspNet.Mvc = net45
Castle.Facilities.AspNet.Mvc.Tests = net45
Castle.Facilities.AspNet.SystemWeb = net45
Castle.Facilities.AspNet.SystemWeb.Tests = net45
Castle.Facilities.AspNet.WebApi = net45
Castle.Facilities.AspNet.WebApi.Tests = net45
Castle.Facilities.AspNetCore = netstandard2.0
Castle.Facilities.AspNetCore.Tests = netcoreapp2.0
Castle.Facilities.Logging = net45; netstandard2.0
Castle.Facilities.WcfIntegration = net45
Castle.Facilities.WcfIntegration.Demo = net45
Castle.Facilities.WcfIntegration.Tests = net45
Castle.Windsor = net45; netstandard2.0
Castle.Windsor.Extensions.DependencyInjection = netstandard2.0
Castle.Windsor.Extensions.DependencyInjection.Tests = net5.0; 
Castle.Windsor.Tests = net45; net5.0
```

Having not done much with the Windsor codebase, is it insane to suggest aligning with the Core TFM list, where reasonable? Does targeting `net6.0` even make sense for Windsor? At the very least, it should support `net462;netstandard2.0` against Castle.Core@5.0.0, which might be a good first step.

Could also consider retiring some of the Facilities, download counts don't always set the world on fire (https://www.nuget.org/packages/Castle.Facilities.AspNet.Mvc/, https://www.nuget.org/packages/Castle.Facilities.AspNet.WebApi/).

Jevonius:
Implements #612, specifically:
* Updates `Castle.Core` references to `5.0.0`
* Changes `net45` target to `net462` 
* Removes obsolete `FEATURE_NETCORE_REFLECTION_API`


Jevonius:
Updated target frameworks:
```
Castle.Facilities.AspNet.Mvc = net462
Castle.Facilities.AspNet.Mvc.Tests = net462
Castle.Facilities.AspNet.SystemWeb = net462
Castle.Facilities.AspNet.SystemWeb.Tests = net462
Castle.Facilities.AspNet.WebApi = net462
Castle.Facilities.AspNet.WebApi.Tests = net462
Castle.Facilities.AspNetCore = netstandard2.0
Castle.Facilities.AspNetCore.Tests = netcoreapp3.1
Castle.Facilities.Logging = net462; netstandard2.0
Castle.Facilities.WcfIntegration = net462
Castle.Facilities.WcfIntegration.Demo = net462
Castle.Facilities.WcfIntegration.Tests = net462
Castle.Windsor = net462; netstandard2.0
Castle.Windsor.Extensions.DependencyInjection = net462; netstandard2.0
Castle.Windsor.Extensions.DependencyInjection.Tests = net462; netcoreapp3.1
Castle.Windsor.Tests = net462; netcoreapp3.1
```

src/Castle.Windsor/Compatibility/PermissionUtil.cs
```diff
@@ -12,20 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
+#if FEATURE_SECURITY_PERMISSIONS
 
-#if NET45
-namespace Castle.Core.Internal
+namespace Castle.Windsor.Compatibility
```
jonorossi, line 17: This doesn't belong in the `Castle.Windsor.Compatibility` namespace which is now empty? Windsor now owns this class and can refactor it however it likes, it isn't a compatibility layer.

It also looks to only be used inside the `Castle.Windsor` assembly so it shouldn't be public.

src/Castle.Windsor.Extensions.DependencyInjection/Castle.Windsor.Extensions.DependencyInjection.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0</TargetFrameworks>
```
jonorossi, line 4: Does `Castle.Windsor.Extensions.DependencyInjection` run on .NET Framework? I thought the MS libraries were .NET Core only.

jonorossi (changes requested):

src/Castle.Windsor/Castle.Windsor.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-    <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0</TargetFrameworks>
```
jonorossi, line 4: Castle.Core is targeting `net462;netstandard2.0;netstandard2.1;net6.0`. Is there any value in this also having net6.0?

src/Castle.Windsor.Extensions.DependencyInjection/Castle.Windsor.Extensions.DependencyInjection.csproj
Jevonius: That’s odd, must have been on autopilot, don’t _think_ this was intentional. Will check later.

Jevonius: Wasn't intentional; reverting.

Jevonius:
I had a fiddle with rebase to fix the erroneous addition of `net462` for `Castle.Windsor.Extensions.DependencyInjection` and have managed to lose your comment. Think I'll stick to an additional commit next time :)

src/Castle.Windsor.Extensions.DependencyInjection/Castle.Windsor.Extensions.DependencyInjection.csproj
Jevonius: Resolved - rebased and confused myself, but have got rid of the addition.

src/Castle.Windsor/Castle.Windsor.csproj
Jevonius: I initially was going to add `net6.0` as part of this issue, but realised it would be better to do them separately. Had a quick throw-away stab at it, and it'll involve some fiddling with package references, but seems pretty straightforward.

src/Castle.Windsor/Compatibility/PermissionUtil.cs
Jevonius: Ok, I'll shift it to the `.Core.Internal` namespace. Given Castle.Core has dropped CAS support, I suspect something similar will happen with Windsor (i.e. this class won't live long), but need to understand it a bit more first.

Jevonius: There's no `InternalsVisibleTo` attribute on Castle.Windsor, so if I mark the class `internal` I'll have to get rid of the test. Happy to go either way, although I assume because the attribute isn't there already, the preference is to just not have a test?

jonorossi: > Ok, I'll shift it to the `.Core.Internal` namespace. Given Castle.Core has dropped CAS support, I suspect something similar will happen with Windsor (i.e. this class won't live long), but need to understand it a bit more first.

Sounds good.

> There's no `InternalsVisibleTo` attribute on Castle.Windsor, so if I mark the class `internal` I'll have to get rid of the test. Happy to go either way, although I assume because the attribute isn't there already, the preference is to just not have a test?

Yer, we wouldn't want `InternalsVisibleTo`, the unit tests should only test the public API. I don't mind whether you stick it in the Internal namespace and mark it public, or delete the test. I'll let you decide.

src/Castle.Windsor.Extensions.DependencyInjection/Castle.Windsor.Extensions.DependencyInjection.csproj
jonorossi: The unit test project is still changed from `net5.0` to `netcoreapp3.1`.

src/Castle.Windsor/Castle.Windsor.csproj
jonorossi: One at a time sounds great, easier for me to review. Obviously need to get the build set up with it too.

src/Castle.Windsor.Extensions.DependencyInjection/Castle.Windsor.Extensions.DependencyInjection.csproj
Jevonius: That's intentional, for two reasons:
1) `net5.0` is out of support
2) `netcoreapp3.1` is preferable to `net6.0`, as otherwise adding `net6.0` to the base project would mean the `netstandard2.0` output is no longer tested.

jonorossi: Thanks for the explanation 👍

src/Castle.Windsor/Compatibility/PermissionUtil.cs
Jevonius: Haven't had a chance to get back into the code for this, but did some reading up on internals and testing. Background: I'm used to having internals visible to test projects (and this is [what Castle.Core does too](https://github.com/castleproject/Core/blob/v5.0.0/src/Castle.Core/Properties/InternalsVisibleToTests.cs)), as otherwise you can end up with an awful lot of internal code that can't easily be tested independently (as it's hidden under other functionality); wanted to check my reasoning.

It's a bit long-winded, but [LosTechies has a good post about it](https://lostechies.com/derickbailey/2014/01/03/semantics-modules-and-testing-why-and-how-i-test-internal-components-not-private-methods/) that aligns with my thoughts. I quite liked the car engine analogy:
> Engines are not “private” components, in my experience. They are “internal” components of the car. They are complex, prone to faults from many parts, and must be rigorously tested prior to mounting them in the vehicle. Failure to do this leads to cars that have faulty engines.

Having said all that... This is one tiny test for some code that will hopefully get deprecated soon, so not really worth worrying about. I'll just drop the test!

Jevonius: Per discussion in #612, this is good to go now - I ended up getting rid of the tests.

jonorossi (approved):

PR #618, opened 2022-07-08
Create parallel data structure for types to fasten up lookups later.

sqeezy:
When profiling application build times I noticed that a lot of my performance went into a List.Contains call.
This seems like a problem that's easily solved. I'm well aware that I might have missed some detail here, but this kind of change would bring a lot of performance to applications registering a lot of components.

This is just a rough draft, but I would be very much interested in your input and explanations.

Thank you!

### DIFF
--- a/src/Castle.Windsor/Core/ComponentModel.cs
+++ b/src/Castle.Windsor/Core/ComponentModel.cs
@@ -40,6 +40,7 @@ namespace Castle.Core
 
 		[DebuggerBrowsable(DebuggerBrowsableState.Never)]
 		private readonly List<Type> services = new List<Type>(4);
+		private HashSet<Type> servicesFast = new HashSet<Type>();
 
 		[DebuggerBrowsable(DebuggerBrowsableState.Never)]
 		private ComponentName componentName;
@@ -331,6 +332,12 @@ namespace Castle.Core
 		{
 			get { return services; }
 		}
+		
+		[DebuggerDisplay("Count = {services.Count}")]
+		public HashSet<Type> FastServices
+		{
+			get { return servicesFast; }
+		}
 
 		[DebuggerBrowsable(DebuggerBrowsableState.Never)]
 		internal ParameterModelCollection ParametersInternal
@@ -378,7 +385,7 @@ namespace Castle.Core
 					              type));
 			}
 
-			ComponentServicesUtil.AddService(services, type);
+			ComponentServicesUtil.AddService(services,servicesFast, type);
 		}
 
 		/// <summary>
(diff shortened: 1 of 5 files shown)


src/Castle.Windsor/Core/ComponentModel.cs
```diff
@@ -331,6 +332,12 @@ public IEnumerable<Type> Services
 		{
 			get { return services; }
 		}
+		
+		[DebuggerDisplay("Count = {services.Count}")]
+		public HashSet<Type> FastServices
+		{
+			get { return servicesFast; }
+		}
```
jonorossi, line 340: This is on the public interface, I don't think that is right. What sort of performance improvement are you seeing?

Could ComponentModel just have a ContainsService (or similar name) method so the internals aren't exposed?

jonorossi (changes requested):

src/Castle.Windsor/Core/ComponentModel.cs
sqeezy: I agree that this should not be part of the public api. I just removed it from there. I just internalized the collection for now. Exposing the collection itself internally seems necessary. For that look into [ComponentServicesUtil](https://github.com/sqeezy/Windsor/blob/faster_type_lookup/src/Castle.Windsor/Core/Internal/ComponentServicesUtil.cs#L25).

The root of my problems is in [AbstractHandler](https://github.com/sqeezy/Windsor/blob/faster_type_lookup/src/Castle.Windsor/MicroKernel/Handlers/AbstractHandler.cs#L186). When registering a lot of components the main effort of our application build goes into the Contains implementation of List<T>, which does not scale well. That effort is reduced to almost none with a parallel HashSet<T> in use.

src/Castle.Windsor/Core/ComponentModel.cs
```diff
@@ -40,6 +40,7 @@ public sealed class ComponentModel : GraphNode
 
 		[DebuggerBrowsable(DebuggerBrowsableState.Never)]
 		private readonly List<Type> services = new List<Type>(4);
+		private HashSet<Type> servicesFast = new HashSet<Type>();
```
jonorossi, line 43:
```suggestion
		private readonly HashSet<Type> servicesLookup = new HashSet<Type>();
```

jonorossi (changes requested):
Makes sense with the util method being used from 2 places.

Naming wise, let's go with the purpose rather than "fast" to keep it consistent.

Watch out for code formatting to kept things consistent.

sqeezy:
All your proposed changes are commited.
I think the formatting did not get changed in the process.

jonorossi (approved):

PR #628, opened 2022-09-15
Split hosting extensions into Castle.Windsor.Extensions.Hosting

### LINKED ISSUE #625: Split hosting extensions from Castle.Windsor.Extensions.DependencyInjection into separate package
`Castle.Windsor.Extensions.DependencyInjection` contains a dependency to hosting related packages (`Microsoft.Extensions.Hosting`). You however don't always need hosting when using `Microsoft.Extensions.DependencyInjection.Abstractions`. It would be great if the hosting related functions could be split off into a `Castle.Windsor.Extensions.Hosting` package.

For example, I'm working on implementing Windsor for a front-end application which uses Microsofts DI abstractions. I'd prefer not to have hosting related dependencies in my front-end application.

#624 would reduce the pain, but the change suggested in this issue would be much better.

ikkentim:
Closes #625


jonorossi (approved):

jonorossi:
Looks great, merged.

PR #641, opened 2022-11-30
Changes to only support .net 7.0 and onwards

renemadsen:
This removes support for all older framworks older than .net 7.0
This also removes tests for those removed frameworks.
This adds:

- Dependabot
- Github Actions build scripts
- Github Actions build script for releasing to nuget.org for faster deployment of new releases.


jonorossi:
Closing this pull request. We won't be dropping all target frameworks prior to .NET 7.

PR #645, opened 2023-01-20
Fixed null access

zii-dmg:
`parentContext` was checked for null after access to `parentContext.Handler` and `.ReleasePolicy`

### DIFF
--- a/src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
+++ b/src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
@@ -67,7 +67,7 @@ namespace Castle.MicroKernel.Context
 		/// <param name = "parentContext"> The parent context. </param>
 		/// <param name = "propagateInlineDependencies"> When set to <c>true</c> will clone <paramref name = "parentContext" /> <see cref = "AdditionalArguments" /> . </param>
 		public CreationContext(Type requestedType, CreationContext parentContext, bool propagateInlineDependencies)
-			: this(parentContext.Handler, parentContext.ReleasePolicy, requestedType, null, null, parentContext)
+			: this(parentContext?.Handler, parentContext?.ReleasePolicy, requestedType, null, null, parentContext)
 		{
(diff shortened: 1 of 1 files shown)


jonorossi:
We really need a unit test for a case where the parent context is null. If this is a bug, it hasn't surfaced in nearly 20 years.

zii-dmg:
Found by static analyzer.

zii-dmg:
It's public API. Anyone can pass null and will get `NullReferenceException` instead of `ArgumentNullException`.

src/Castle.Windsor/MicroKernel/Context/CreationContext.cs
```diff
@@ -67,7 +67,7 @@ public class CreationContext :
 		/// <param name = "parentContext"> The parent context. </param>
 		/// <param name = "propagateInlineDependencies"> When set to <c>true</c> will clone <paramref name = "parentContext" /> <see cref = "AdditionalArguments" /> . </param>
 		public CreationContext(Type requestedType, CreationContext parentContext, bool propagateInlineDependencies)
-			: this(parentContext.Handler, parentContext.ReleasePolicy, requestedType, null, null, parentContext)
+			: this(parentContext?.Handler, parentContext?.ReleasePolicy, requestedType, null, null, parentContext)
```
jonorossi, line 70: The problem with doing this is that you call the other constructor passing a null `IHandler` and `IReleasePolicy` which isn't valid.

You'll find that Windsor likely has a bunch of these "bugs" that aren't hurting anyone because Windsor's original design was to be completely open with everything marked public, you can see `ResolutionContext` in the same file is public with no checking, but is never created by users. We've been slowly closing things with Castle DynamicProxy because we needed to to keep the public API sane and make changes, however Windsor is still completely open.

I'd like to see Windsor enable nullable reference types, however accepting this pull request would mean that the other constructor needs to accept null and run to completion to then run this constructor which throws.

jonorossi (changes requested):

PR #655, opened 2023-07-20
Add GitHub actions

Toxantron:
I took the relevant changes from #641 about GitHub actions and combined them with the [actions from Castle Core](https://github.com/castleproject/Core/blob/master/.github/workflows/build.yml).

Right now I am having an issue with exit code one caused by test projects not availabe for a certain framework version.


.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,85 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    
+    # Install all supported .NET versions
+    - name: Install .NET Core 3.1
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.x
```
jonorossi, line 22: I'd say it's time to remove support for .NET Core 3.1, support ended 2022-12-13. We've got .NET 6 targeting in 6.0.0. If 3.1 gets in the way, submit a separate PR for it.

jonorossi:
@Toxantron I've marked your pull request draft. I've added a Gmail filter to star Castle tickets so they don't get lost in my inbox, but mention me if I don't respond in a couple of days.

.github/workflows/dotnet.yml
jonorossi: I don't have an objection if .NET Framework was removed but .NET Standard 2.0 was kept. That means users that still haven't migrated have an easier path by keeping support for .NET Framework 4.6.1+ (really .NET Framework 4.7.2) and means .NET Framework only technologies/APIs can be removed. Maybe there is no need for .NET Standard either, however check out some open source libraries that depend on Windsor to find out their targets.

Can you create an issue describing exactly what you propose and ask for others to provide their opinions/proposals.

Toxantron: For now I switched the netcore3.1 test to net 6 and got the net462 tests running as well. I will create a separate issue for discussing end of support for older frameworks.

Toxantron:
@jonorossi I completed the action scripts for Windows and Linux. On Linux we run everything for net6.0 and on windows the net462 tests only. The successful runs and unit tests you can find [here](https://github.com/PHOENIXCONTACT/Windsor/actions).

I kept the release script from @renemadsen and cleaned it up a little bit, but it is basically untested.

Let me know, what you think.

Toxantron:
@jonorossi do you need anything else to merge this? I could test the release script under a different ID on our myget feed, still would not garantuee correct behavior for upstream.

With this being merged we could at least disable AppVeyor CI runs.

.github/workflows/dotnet-release.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Get the version
+      id: get_version
+      run: echo ::set-output name=VERSION::$(echo $GITHUB_REF | cut -d / -f 3 | cut -d "v" -f 2)
+
+    - name: Build
+      run: dotnet build Castle.Windsor.sln --configuration Release
+    - name: Pack
+      run: dotnet pack Castle.Windsor.sln -c Release -o artifacts -p:PackageVersion=${{ steps.get_version.outputs.VERSION }}
+
+    - name: Push generated Rebus package to GitHub registry
```
jonorossi, line 42: This isn't Rebus and it's not being pushed to GitHub Packages.

We have a NuGet feed with AppVeyor for all builds, can we do the same with GitHub Packages so we can test the packages.

.github/workflows/dotnet-release.yml
```diff
@@ -0,0 +1,43 @@
...
+  build-lin:
+    name: Build and test Linux
+    uses: ./.github/workflows/build-test-linux.yml
+      
+  build-win:
+    name: Build and test on Windows
+    uses: ./.github/workflows/build-test-win.yml
+      
+  deploy:
+    needs: [build-lin, build-win]
```
jonorossi, line 18: Why does the deployment need both Linux and Windows build binaries?

jonorossi:
I've added a couple review comments. If you can get packages published to GitHub Packages I'll take a detailed look at the output to confirm they are what I expect.

.github/workflows/dotnet-release.yml
Toxantron: I wanted to make sure that build and deployment run build and test again before publishing. In our projects it is standard practise to have CI runs with tests and still perform them again in the release pipeline.

jonorossi: Fair enough. However, the deployment pipeline doesn't actually use the tested binaries.

Toxantron: No it does not. Uploading and downloading artifacts did not seem worth the effort. I only use it as a condition.

Toxantron:
I confirmed the [release pipeline](https://github.com/PHOENIXCONTACT/Windsor/actions/runs/5679327222) works and I have created [packages](https://github.com/orgs/PHOENIXCONTACT/packages?repo_name=Windsor) for tags.

However using `GITHUB_TOKEN` causes issues when pushing multiple packages for the first time. Rerunning the same job creates the missing packages. Since we rebuild anyways this does not even cause issues with expired artifacts.

jonorossi:
The AppVeyor package feed is for CI builds (i.e. every build), you don't have to tag them. See the readme:
https://github.com/castleproject/Windsor#nuget-preview-feed

Maybe we don't really need it anymore, and we should just tag pre-release packages more often publishing betas to nuget.org.

Since you've got that publishing working, leave it for now. I don't have time tonight too look at the package contents, but it's a great test for publishing to nuget.org.

Toxantron:
We could have both. Push CI packages to GitHub after every merge on main and push to nuget.org for tags.

Toxantron:
A shame we never got this across the finish line.

Issue #595, opened 2021-04-05
 Castle.Windsor.Extensions.DependencyInjection No Scope Available


EricMKaufman:
I am using this nuget package to configure Castle Windsor as the DI container. https://www.nuget.org/packages/Castle.Windsor.Extensions.DependencyInjection/

This works fine except for when I try to resolve classes with an IHttpHandler. I get the error: InvalidOperationException: No scope available

What is the correct way to create scope for an HttpHandler?

Here's a sample application reproducing the issue:

```c#
namespace WebApplication1
{
    public class Global : System.Web.HttpApplication
    {
        public static IWindsorContainer Container;

        protected void Application_Start(object sender, EventArgs e)
        {
            Console.WriteLine("Hello World!");

            Container = new WindsorContainer();
            var host = new HostBuilder()
                .UseWindsorContainerServiceProvider(Container)
                .ConfigureServices((hostContext, services) =>
                {
                    services.AddScoped<IFakeHttpClient, FakeHttpClient>();
                    services.AddHttpClient<IFakeHttpClient, FakeHttpClient>((client) =>
                    {
                        client.BaseAddress = new Uri("http://test");
                    });
                })
                .Build();

            //Note: This works great
            var resolved = Container.Resolve<IFakeHttpClient>();
        }

        protected void Session_Start(object sender, EventArgs e)
        {

        }

        protected void Application_BeginRequest(object sender, EventArgs e)
        {

        }

        protected void Application_AuthenticateRequest(object sender, EventArgs e)
        {

        }

        protected void Application_Error(object sender, EventArgs e)
        {

        }

        protected void Session_End(object sender, EventArgs e)
        {

        }

        protected void Application_End(object sender, EventArgs e)
        {

        }
    }


    public class FakeHttpClient : IFakeHttpClient
    {
        private readonly HttpClient _httpClient;

        public FakeHttpClient(HttpClient httpClient)
        {
            _httpClient = httpClient;
        }
    }

    public interface IFakeHttpClient
    {
    }
    public class HttpHandler : IHttpHandler
    {
        public void ProcessRequest(HttpContext context)
        {
            using (var scope = Global.Container.BeginScope())
            {
                //error here
                var httpClient = Global.Container.Resolve<IFakeHttpClient>();
            }
        }

        public bool IsReusable => false;
    }


}
```

EricMKaufman:
[...]

PR #662, opened 2023-08-06
Bugfix ms di no available scope

### LINKED ISSUE #646: NET 6.0  ForcedScope Failed with "No Available Scope" on  GET request
First of all, great work all of you. I'm trying to port some project of mine to NET 6.0 and AspNetCore, which are new to me. 
I've downloaded master, and using the Castle.Windsor.Extensions.Dependency.Injection. 
When I run a simple Hello World I'm facing a "No Scope Available exception".

My initialization code is pretty simple, just trying to put all together:

    private static void Main(string[] args)
    {
        IApplicationInitializer initializer;
        var builder = WebApplication.CreateBuilder(args);
        builder.Host.UseTaxologicCoreInitialization("WebPortal", out initializer);
        initializer.Initialize();

        var app = builder.Build();

        app.MapGet("/", () => "Hello World!");

        app.Run();
    }

IAplicationInitializer implementation does all the windsor registration.

UseTaxologicCoreInitialization  is just

        public static IHostBuilder UseTaxologicCoreInitialization(this IHostBuilder hostBuilder, string _webSiteName, out IApplicationInitializer initializer)
        {
            var ioc = new WindsorContainer();
            var f = new WindsorServiceProviderFactory(ioc);
            ioc.Register(Component.For<IHostBuilder>().Instance(hostBuilder));

            initializer = new WindsorApplicationInitializer(ioc, new AppPathFinder(), _webSiteName);

            return hostBuilder.UseWindsorContainerServiceProvider(f);
        }

If I run this code, it fails at app.Run() when it gets it first request to /  with an exception "No Scope Available" at line 27 of ExtensionContainerScopeCache , 
that comes fron ForcedScope constructor.

For what I can infere from the code, as I'm no expert in Castle internals, the ForcedScope stashes the current Scope, set as current the passed one and restores it on ForcedSope disposal. The problem is that when there is no current scope in the cache it fails when tries to get it to set previousScope in
		
```
internal ForcedScope(ExtensionContainerScopeBase scope)
		{
			previousScope = ExtensionContainerScopeCache.Current;
			this.scope = scope;
			ExtensionContainerScopeCache.Current = scope;
		}
```

I did a couple of  small changes to prevent that,  and the exception has gone.
These are the changes so you can evaluate them and include if you think that has value:

in ExtensionContainerScopeCache I added a getter to know if Current has value

```
internal static bool hasContext
		{
			get => current.Value is not null;
		}

```
in ForcedScope I changed the Constructor with this
		
```
internal ForcedScope(ExtensionContainerScopeBase scope)
		{
			if (ExtensionContainerScopeCache.hasContext)
				previousScope = ExtensionContainerScopeCache.Current;
			else
				previousScope = null;
			this.scope = scope;
			ExtensionContainerScopeCache.Current = scope;
		}
```

It just prevent the exception if there is no current scope

Looking forward to hear your comments about the changes, or any comments on how to prevent the exception by other means

Thanks in advance

Fernando

rvdginste:
Fixes #646


AGiorgetti:
I tested this code and there are cases in which it still fail in an AspNet Core project.

The main issue is:
- register a singleton service in the service collection (like a logger).
- start an Unsafe threadpoool thread
- resolve the service using the underlying WindsorContainer (other libraries might have their own wrapper around castle) leads to a null reference exception in the lifestyle scope accessors.

I openend another PR with some more tests and code changes to be evaluated that can help finding a solution

rvdginste:
@AGiorgetti @jonorossi 

I am starting to think that the safest thing to do now (at least for now) is to revert the code to the original static root sope, so that the code is 100% compatible with the earlier version of the extension. Then we no longer support concurrent containers, but existing code should not have problems and it would allow existing projects to upgrade to the new version.

I do want to explore this further, but I am afraid it will take some time to have a good solution.

What do you think?

Ruben

jonorossi:
@rvdginste I'm happy to leave that decision to you guys who are using it.

My only request is to keep all the additional unit tests you are writing, and if you plan on restoring concurrent containers in the future keep them separate and mark `[Ignore("...")]`.

/cc @AGiorgetti

AGiorgetti:
I have no problem reverting the code as long as we keep all the tests (I added even more checking what happens when you attempt to resolve services with different lifecycles in another PR #664)

Looking at the code of dotnet DI you can see that they use a root scope tied to each ServiceProvider they create from the factory; some of the thing we should investigate to support "concurrent" containers (they are not anyway, because the underlying instance of castle windsor is always the same) - imo - are:

- create a "root scope" for each IServiceProvider resolved from the factory: allow for multiple rootscope and create them every time the call CreateServiceProvider() from the provider factory, this should be more similar to what dotnet does.
  We should also check the Dispose() method of WindsorScopedServiceProvider: it should not dispose the container.
- experiment with child containers to simulate dotnet scopes.

Another thing to investigate is how Singleton are handled, I'm not sure having a custom NetStatic lifecycle is a good idea given the fact that under the hood we have a single WindsorContainer instance... why not use the standard Singleton behavior?

anslmorvan:
For some reason I'm not aware of, my code that used to work with previous versions does not work with the latest. I have added a comment in #595 but I think it's worth having it here.

eekodeerder:
My apologies, I am not following the conversation that well.  I have ported a set of applications to .Net 6.0 from .net core 3.0.  All of them use Castle containers.  Some of them use it as a web service on MVC in asp.net.
I am getting the "System.InvalidOperationException: No scope available
   at Castle.Windsor.Extensions.DependencyInjection.Scope.ExtensionContainerScopeCache.get_Current()
   at Castle.Windsor.Extensions.DependencyInjection.WindsorScopedServiceProvider.GetRequiredService(Type serviceType).." exception.
1.  I cannot ascertain:
  a.  Is there a fix for this bug in  6.0.0?
  b.  Is there an indication when this fix (if it is fixed) will be released?
2.  Is there a workaround for this problem using the .net 3.0 method of starting up/configuring the container, or is there a fix which uses a different way of configuring the container?
3. Is there a way for me to obtain the .pdb (symbol) files for the released version to help me identify where it breaks, trying to find a work-around.
4.  Is there a way in which I can contribute to a fix?  This is a rather nasty breakage at the moment for me.

PR #663, opened 2023-08-10
RFC: Replace License URL with expression

### LINKED ISSUE #657: NuGet packages still using deprecated `licenseUrl` element
> WARNING: The <licenseUrl> element is deprecated. Consider using the <license> element instead. Learn more: https://aka.ms/deprecateLicenseUrl.,Readme missing. Go to https://aka.ms/nuget-include-readme learn How to include a readme file within the package.

Could someone submit a pull request to resolve this. The build warning came before tooling supported the newer `license` element so we suppressed the warning (#481), that suppression can now be removed too.

jeme:
This is just an initial attempt to fix: #657 seeking comments on the approach.

 - Replaced the License URL with the SPDX License expression for Apache 2.0 (see https://spdx.org/licenses/)
 - Add Readme file to packages

The following files is currently added to packages as well:
 - ..\..\LICENSE
 - ..\..\buildscripts\ASL - Apache Software Foundation License.txt
 
 This seems a bit redundant, I think it's fair to bundle the actual license with the package, but shouldn't it just be one of those, and preferable the one in the root?


jonorossi (changes requested):
Thanks @jeme.

Regarding the LICENSE file, I agree we should only have one and it should be in the root. Historical mess.

Can you replace `LICENSE` with the content from `buildscripts\ASL - Apache Software Foundation License.txt` so we are including the full license. GitHub appears to only detect the full license text rather than just the file header.

* https://github.com/apache/kafka/blob/trunk/LICENSE
* https://github.com/apache/spark/blob/master/LICENSE

Don't forget to remove the `licenseUrl` warning suppression as mentioned in the ticket.

jonorossi:
@jeme Can you also do the replacement as 2 commits so git reports a rename without content change in the second commit.

jeme:
> @jeme Can you also do the replacement as 2 commits so git reports a rename without content change in the second commit.

Will try to get it done that way, it will likely require me fiddling a bit around.

jeme:
Forgot the part about suppression, should be fixed now.

As for the 2 commit thing, I had to delete ./LICENSE in the first commit before moving the other file into that location in the second commit to get it to recognize it as a rename as far as I could figure out.

It TECHNICALLY leaves the first commit in a broken state, while I could technically fix that up, for now I didn't as it would generate extra noise instead as I would have to change the common.props in both commits.

If you wish I do that just say.

jonorossi (approved):
You've done it exactly as I expected. I just wanted the history of the LICENSE file to show a rename so it's clear nothing has actually changed.

jonorossi:
GitHub license detection is fixed as expected:

Screenshot 2023-08-10 at 9 26 32 pm

PR #666, opened 2024-01-18
Bugfix net6 ms di no available scope #662

weixiang-li:


morindo:
This fix works on my machine! ❤️
.NET 6.0

eekodeerder:
This fix works on my projects.  I can now use net6.0.  Thank you.😍

SandraMarszalek:
If it works as other users said can we please get a new release including this change? It is a great blocker

PR #668, opened 2024-02-15
Update Castle.Windsor.Extensions.DependencyInjection to support .NET8

alkampfergit:
This Pull Request contains also all modification by Alessandro Giorgetti already made in PR #664 related to issue #646 

This Pull Request aim to give Castle.Windsor.Extensions.DependencyInjection full support to .NET 8 new DI model with keyed service. The reason of this Fork was the inability to upgrade ours production code to .NET 8 due to internal error in Castle.Windsor. This problem happens especially with Microsoft Orleans 8 and sometimes code running on Kestrel 8. 

1. Modified code to support KeyedService in .NET 8. All basic test from framework passes, also this pr contains some other tests that idenfy some problem that occurred in our production code
2. Scope management: Orleans 8 changed how it resolves grains so we always got a null Root Scope. This happens because the Root Scope is marked with AsyncLocal, but Orleans uses thread differently so that AsyncLocal is always null. 
3. We introduced a modification in base Castle.Windsor to allow for IHandler to return the current kernel object. While this is somewhat an Hack, it was the only way for us to find the correct root scope when that AsyncLocal is null. This Hack can be avoided if we always use one single instance container, but this is a too restrictive situation.

We actually start using this code in our production code, because without these modification we have as only option to use Orleans 8 to entirely remove Castle.Windsor from our project (An extemely big work) so this code, starting from today, is running in our solution.

Feel free to ask any information or require modification in code if you feel that something is wrong.

Thanks.

PR #670, opened 2024-05-19
Use Root scope when current is null (fixes #646)

### LINKED ISSUE #646: NET 6.0  ForcedScope Failed with "No Available Scope" on  GET request

ltines:
Fixes #646 by using Root scope when current scope in `ExtensionContainerScopeCache.current` is `null`. This can happen when `WindsorScopedServiceProvider` is called from a new thread (or rather different async context).

In this situation `ForcedScope` uses the original root scope. It also restores it as `ExtensionContainerScopeCache.current` on dispose.

The only case that I haven't figured out is when resolving directly via `IWindsorContainer` (and not through `IServiceProvider`) on a new thread. `ExtensionContainerScopeAccessor.GetScope` still throws "No scope available"

### DIFF
--- a/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScopeCache.cs
+++ b/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScopeCache.cs
@@ -19,7 +19,7 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Scope
 
 	internal static class ExtensionContainerScopeCache
 	{
-		internal static readonly AsyncLocal<ExtensionContainerScopeBase> current = new AsyncLocal<ExtensionContainerScopeBase>();
+		private static readonly AsyncLocal<ExtensionContainerScopeBase> current = new AsyncLocal<ExtensionContainerScopeBase>();
 		/// <summary>Current scope for the thread. Initial scope will be set when calling BeginRootScope from a ExtensionContainerRootScope instance.</summary>
 		/// <exception cref="InvalidOperationException">Thrown when there is no scope available.</exception>
 		internal static ExtensionContainerScopeBase Current
@@ -27,5 +27,7 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Scope
 			get => current.Value ?? throw new InvalidOperationException("No scope available");
 			set => current.Value = value;
 		}
+
+		internal static ExtensionContainerScopeBase TryCurrent => current.Value;
 	}
 }
--- a/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
+++ b/src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScope.cs
@@ -35,9 +35,9 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Scope
 
 		public override void Dispose()
 		{
-			if (ExtensionContainerScopeCache.current.Value == this)
+			if (ExtensionContainerScopeCache.TryCurrent == this)
 			{
-				ExtensionContainerScopeCache.current.Value = parent;
+				ExtensionContainerScopeCache.Current = parent;
 			}
 			base.Dispose();
 		}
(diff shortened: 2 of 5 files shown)


alkampfergit:
I still ask to check for our pr #668 that will fix this and also add support for .NET 8 keyed services. If it is possible could you check code of pr #668 to verify if this fixes null scope in your situation? I'm asking because the support for .NET 8 branch is opened from a long time and I'd like it to be closed if possible (we are in production with that version using a private package source and I'd really like to move again on supported nuget packages)

ltines:
@alkampfergit i had a quick look, but need to read a lot about semantics (especially keyed / not keyed). I would like to implement it using named instances in Castle so we don't have to introduce new concepts.

alkampfergit:
Remember that if you are moving to dependency injection version 8.0.0 without keyed service support the adapter does not works . The problem presents when any library starts using keyed service and new resolution semantic, so you got weird error.

src/Castle.Windsor.Extensions.DependencyInjection/Scope/ExtensionContainerScopeCache.cs
```diff
@@ -19,13 +19,15 @@ namespace Castle.Windsor.Extensions.DependencyInjection.Scope
...
+		private static readonly AsyncLocal<ExtensionContainerScopeBase> current = new AsyncLocal<ExtensionContainerScopeBase>();
 		/// <summary>Current scope for the thread. Initial scope will be set when calling BeginRootScope from a ExtensionContainerRootScope instance.</summary>
 		/// <exception cref="InvalidOperationException">Thrown when there is no scope available.</exception>
 		internal static ExtensionContainerScopeBase Current
 		{
 			get => current.Value ?? throw new InvalidOperationException("No scope available");
 			set => current.Value = value;
 		}
+
+		internal static ExtensionContainerScopeBase TryCurrent => current.Value;
```
jonorossi, line 31: `TryCurrent` is a strange name, it's not a function. Any reason not to just name it `Current` and just document that is can return null?

jonorossi (changes requested):
Could you add at least one unit test for this change.

